PR #62, opened 2022-08-02
Feature/controls 1166

TMCamp:
*What does this PR do?*
Adds Catalog version and Impact level to Catalog model.
Adds catalog_version to Project model.
Extends save() method to add Catalog based on Project catalog version and impact level.
Creates Control model to store Project specific Control data.

*Jira ticket number?*

*Test*
- [ ] Run migrations to create the new table
- [ ] Using Postman:
   - Create a Catalog:
     - Fields _name_, _file_name_, _impact_level_; the file_name should reference a [valid ARS 3.1 catalog](https://github.com/CMSgov/ars-machine-readable/blob/main/3.1/oscal/json/CMS_ARS_3_1_LOW-baseline_catalog.json)
- [ ] In the UI create a Project
- [ ] Check your controls in the API by going to http://localhost:8000/api/projects/[PROJECT ID]/controls
- [ ] You should see a paginated list of controls.


projects/models.py
```diff
@@ -159,3 +178,48 @@ def add_components_for_project(sender, instance, **kwargs):
...
+        NOT_STARTED = 1, "Not started"
+        INCOMPLETE = 2, "Incomplete"
+        COMPLETE = 3, "Complete"
+
+    class Responsibility(models.IntegerChoices):
+        ALLOCATED = 1, "Allocated"
+        SHARED = 2, "Shared"
+        INHERITED = 3, "Inherited"
+
+    project = models.ForeignKey(Project, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
TMCamp: Yes, the Controls in that table are Project specific, so I think that it would be best to delete them otherwise we'll end up with a bunch of orphaned data.

projects/models.py
```diff
@@ -159,3 +178,48 @@ def add_components_for_project(sender, instance, **kwargs):
...
+    control_id = models.CharField(
+        max_length=12,
+        unique=False,
```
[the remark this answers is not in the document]
TMCamp: These controls are per project so the control_id, for example `ac-1`, will be used in pretty much every Project.

projects/models.py
```diff
@@ -87,15 +93,28 @@ class Project(models.Model):
...
         permissions = [
             ("can_add_members", "Can add members"),
             ("can_delete_members", "Can delete members"),
             manage_project_users_permission,
         ]
 
+    def save(self, *args, **kwargs):
+        version = getattr(self, "catalog_version")
+        impact = getattr(self, "impact_level")
+        if version and impact:
```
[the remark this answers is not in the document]
TMCamp: We will address this in [ISPGBSS-1204](https://jiraent.cms.gov/browse/ISPGBSS-1204)

projects/models.py
```diff
@@ -159,3 +178,48 @@ def add_components_for_project(sender, instance, **kwargs):
...
+        SHARED = 2, "Shared"
+        INHERITED = 3, "Inherited"
+
+    project = models.ForeignKey(Project, on_delete=models.CASCADE)
+    control_id = models.CharField(
+        max_length=12,
+        unique=False,
+        blank=False,
+    )
+    description = models.JSONField(
```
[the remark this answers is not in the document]
TMCamp: The value comes from the OSCAL Catalog which we validate when it is submitted.

.gitignore
```diff
@@ -114,3 +114,6 @@ media/
 # **/migrations/**
 # !**/migrations
 # !**/migrations/__init__.py
+
+# Jetbrains configuration file.
+.idea
```
jwhite3, line 119: Good call. When this gets added to a repo on accident it can cause mayhem 😆

components/models.py
```diff
@@ -73,6 +73,9 @@ class Status(models.IntegerChoices):
     def __str__(self):
         return self.title
 
+    def get_status(self):
+        return self.Status(self.status).label
```
jwhite3, line 77: I'm not entirely sure why this method is necessary, but I might be missing some context. What was your thought here?

components/tests.py
```diff
@@ -80,6 +80,8 @@ def setUpTestData(cls):
             cls.test_catalog = Catalog.objects.create(
                 name="NIST Test Catalog",
                 file_name=catalog,
+                version="NIST 800-53r5",
+                impact_level="moderate",
```
jwhite3, line 84: Future item: Looks like this test catalog is created a bunch of times. We might consider creating a mixin or "base" test case that creates this catalog for different `TestCase`s so that we don't have so many duplicates. (This makes me miss pytest's fixtures 😭)

projects/filters.py
```diff
@@ -0,0 +1,12 @@
+import django_filters
```
jwhite3, line 1: Use `django_filters.rest_framework` instead
```python
from django_filters import rest_framework as filters
```

There are some extended features that will come in handy on the views.

projects/filters.py
```diff
@@ -0,0 +1,12 @@
...
+
+from .models import Control
+
+
+class ControlsFilter(django_filters.FilterSet):
+    status = django_filters.AllValuesMultipleFilter(lookup_expr="iexact")
+    responsibility = django_filters.AllValuesMultipleFilter(lookup_expr="iexact")
```
jwhite3, line 8: Why the need for `AllValuesMultipleFilter` here? From our chat the other day it seemed like maybe a simple `in` would work, but I could be misremembering.

jauschalley (approved):
Praise: Thanks for taking this on thoughtfully.  I know it required a lot of time and effort.  Also, thanks to @jwhite3 for all of your input.

projects/models.py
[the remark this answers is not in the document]
jwhite3: Why did you resolve this? It looks like this comment wasn't addressed.

projects/models.py
```diff
@@ -48,54 +65,55 @@ class Project(models.Model):
         related_name="projects_catalog",
         help_text="Catalog id that this project uses",
     )
-    IMPACT_LEVEL_CHOICES = [
-        ("low", "Low"),
-        ("moderate", "Moderate"),
-        ("high", "High"),
-        ("pii/phi", "PII or PHI"),
-    ]
+    catalog_version = models.CharField(
```
jwhite3, line 68: (chatted about this in person)

This is more of a workflow issue. We need to be able to determine the catalog from inputs to a create project view (including a set of catalog versions), but that would require updating/refactoring the projects list view and whatnot. May be out of scope for this PR though. We could start with this and then clean it up later.

projects/models.py
```diff
@@ -48,54 +65,55 @@ class Project(models.Model):
...
+        max_length=64,
+        default="CMS ARS 3.1",
+        blank=False,
+        help_text="The Catalog version, for example, NIST 800-53r5",
+    )
     impact_level = models.CharField(
```
jwhite3, line 74: Should we add a default `impact_level` that should correspond with the default catalog?

projects/models.py
```diff
@@ -48,54 +65,55 @@ class Project(models.Model):
...
             ("can_add_members", "Can add members"),
             ("can_delete_members", "Can delete members"),
             manage_project_users_permission,
         ]
 
+    def save(self, *args, **kwargs):
+        version = getattr(self, "catalog_version")
+        impact = getattr(self, "impact_level")
+        if version and impact:
+            catalog = Catalog.objects.get(version=version, impact_level=impact)
```
jwhite3, line 113: Need to handle the cases where multiple objects are returned or a matching object isn't found. Otherwise this will result in a 500 with an unhelpful message

projects/models.py
```diff
@@ -159,3 +177,91 @@ def add_components_for_project(sender, instance, **kwargs):
                 instance.components.add(aws_component)
         except ObjectDoesNotExist as e:
             logger.warning(f"Inherited components not found: {e}")
+
+
+@receiver(post_save, sender=Project)
+def add_controls(sender, instance, **kwargs):
```
jwhite3, line 183: (chatted about this in person)

Looking into other options for this workflow.

projects/models.py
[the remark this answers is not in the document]
jwhite3: From our conversation about alternate workflow, we may need to reevaluate this.

projects/models.py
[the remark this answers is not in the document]
jwhite3: Also might need to be reevaluated for the same reason as the cascade delete.

projects/views.py
```diff
@@ -114,6 +116,35 @@ def post(self, request, *args, **kwargs):
         return Response(response, status=status.HTTP_200_OK)
 
 
+class ProjectGetControlList(APIView):
```
jwhite3, line 119: This should probably be a `generics.ListAPIView`

PR #60, opened 2022-08-01
Feature/1178 authenticate apis

jwhite3:
This PR adds basic token authentication to all api views. In addition, a missing dependency (used in `Environment.py`) was added to the dependency list.

*Jira ticket number?*
[ISPGBSS-1178](https://jiraent.cms.gov/browse/ISPGBSS-1178)

*Changes*

- Update User model to include permissions mixin
- Update settings with DRF global default authentication and permissions classes
- Use custom `ObtainAuthToken` view to pass additional user info to the front end

PR #66, opened 2022-08-04
Feature/1178 authenticate apis

jwhite3:
*What does this PR do?*
#60 

Blocked by: 
- [ ] https://github.com/CMSgov/blueprint_ui/pull/62

### DIFF
--- a/blueprintapi/authentication.py
+++ b/blueprintapi/authentication.py
@@ -1,0 +1,30 @@
+import pytz
+
+from datetime import datetime, timedelta
+from django.conf import settings
+from django.utils.translation import gettext_lazy as _
+from rest_framework.authentication import TokenAuthentication
+from rest_framework.exceptions import AuthenticationFailed
+
+
+class ExpiringTokenAuthentication(TokenAuthentication):
+    """Extend TokenAuthentication to expire existing tokens."""
+    def authenticate_credentials(self, key):
+        model = self.get_model()
+        try:
+            token = model.objects.select_related('user').get(key=key)
+        except model.DoesNotExist:
+            raise AuthenticationFailed(_('Invalid token.'))
+
+        if not token.user.is_active:
+            raise AuthenticationFailed(_('User inactive or deleted.'))
+
+        # This is required for the time comparison
+        utc_now = datetime.utcnow()
+        utc_now = utc_now.replace(tzinfo=pytz.utc)
+
+        if utc_now - token.created > timedelta(hours=settings.AUTH_TOKEN_TTL):
+            token.delete()
+            raise AuthenticationFailed(_('Token has expired.'))
+
+        return token.user, token
--- a/blueprintapi/permissions.py
+++ b/blueprintapi/permissions.py
@@ -1,0 +1,13 @@
+from rest_framework.permissions import DjangoObjectPermissions
+
+
+class StrictDjangoObjectPermissions(DjangoObjectPermissions):
+    perms_map = {
+        'GET': ['%(app_label)s.view_%(model_name)s'],
+        'OPTIONS': ['%(app_label)s.view_%(model_name)s'],
+        'HEAD': ['%(app_label)s.view_%(model_name)s'],
+        'POST': ['%(app_label)s.add_%(model_name)s'],
+        'PUT': ['%(app_label)s.change_%(model_name)s'],
+        'PATCH': ['%(app_label)s.change_%(model_name)s'],
+        'DELETE': ['%(app_label)s.delete_%(model_name)s'],
+    }
--- a/blueprintapi/permissions.py
+++ b/blueprintapi/permissions.py
@@ -2,6 +2,7 @@ from rest_framework.permissions import DjangoObjectPermissions
 
 
 class StrictDjangoObjectPermissions(DjangoObjectPermissions):
+    """Extends DjangoObjectPermissions to include GET, OPTIONS, and HEAD permissions requirements."""
     perms_map = {
         'GET': ['%(app_label)s.view_%(model_name)s'],
         'OPTIONS': ['%(app_label)s.view_%(model_name)s'],
--- a/blueprintapi/settings.py
+++ b/blueprintapi/settings.py
@@ -121,6 +121,7 @@ DATABASES = {
 }
 
 AUTH_USER_MODEL = "users.User"
+AUTH_TOKEN_TTL = 24  # Hours
 
 # Password validation
 # https://docs.djangoproject.com/en/3.2/ref/settings/#auth-password-validators
@@ -187,7 +188,7 @@ CORS_ALLOW_HEADERS = [
 
 REST_FRAMEWORK = {
     'DEFAULT_AUTHENTICATION_CLASSES': [
-        'rest_framework.authentication.TokenAuthentication',
+        'blueprintapi.authentication.ExpiringTokenAuthentication',
     ],
     'DEFAULT_PERMISSION_CLASSES': [
         'blueprintapi.permissions.StrictDjangoObjectPermissions',
--- a/blueprintapi/settings.py
+++ b/blueprintapi/settings.py
@@ -190,7 +190,7 @@ REST_FRAMEWORK = {
         'rest_framework.authentication.TokenAuthentication',
     ],
     'DEFAULT_PERMISSION_CLASSES': [
-        'rest_framework.permissions.IsAuthenticated',
+        'blueprintapi.permissions.StrictDjangoObjectPermissions',
     ],
     "DEFAULT_SCHEMA_CLASS": "rest_framework.schemas.coreapi.AutoSchema",
 }
--- a/blueprintapi/settings.py
+++ b/blueprintapi/settings.py
@@ -82,8 +82,6 @@ MIDDLEWARE = [
 
 ROOT_URLCONF = "blueprintapi.urls"
 
-REST_FRAMEWORK = {"DEFAULT_SCHEMA_CLASS": "rest_framework.schemas.coreapi.AutoSchema"}
-
 TEMPLATES = [
     {
         "BACKEND": "django.template.backends.django.DjangoTemplates",
@@ -194,4 +192,5 @@ REST_FRAMEWORK = {
     'DEFAULT_PERMISSION_CLASSES': [
         'rest_framework.permissions.IsAuthenticated',
     ],
+    "DEFAULT_SCHEMA_CLASS": "rest_framework.schemas.coreapi.AutoSchema",
 }
--- a/blueprintapi/settings.py
+++ b/blueprintapi/settings.py
@@ -58,6 +58,7 @@ INSTALLED_APPS = [
     "django_filters",
     "corsheaders",
     "rest_framework",
+    "rest_framework.authtoken",
     "catalogs.apps.CatalogConfig",
     "guardian",
     "components.apps.ComponentsConfig",
@@ -185,3 +186,12 @@ CORS_ALLOW_HEADERS = [
     "x-csrftoken",
     "x-requested-with",
 ]
+
+REST_FRAMEWORK = {
+    'DEFAULT_AUTHENTICATION_CLASSES': [
+        'rest_framework.authentication.TokenAuthentication',
+    ],
+    'DEFAULT_PERMISSION_CLASSES': [
+        'rest_framework.permissions.IsAuthenticated',
+    ],
+}
--- a/blueprintapi/tests.py
+++ b/blueprintapi/tests.py
@@ -1,0 +1,12 @@
+from rest_framework.test import APITestCase
+
+
+class UnauthenticatedAPITestCase(APITestCase):
+    def test_unauthenticated_request_returns_403(self):
+        test_cases = ('/api/projects/', '/api/components/', '/api/catalogs/', '/api/users/')
+
+        for test_case in test_cases:
+            with self.subTest(url=test_case):
+                response = self.client.get(test_case)
+
+                self.assertEqual(response.status_code, 401)
--- a/blueprintapi/tests.py
+++ b/blueprintapi/tests.py
@@ -3,7 +3,7 @@ from rest_framework.test import APITestCase
 
 class UnauthenticatedAPITestCase(APITestCase):
     def test_unauthenticated_request_returns_403(self):
-        test_cases = ('/api/projects/', '/api/components/', '/api/catalogs/', '/api/users/')
+        test_cases = ('/api/projects/', '/api/users/')
 
         for test_case in test_cases:
             with self.subTest(url=test_case):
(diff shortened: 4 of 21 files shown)


blueprintapi/urls.py
```diff
@@ -34,6 +35,7 @@
 urlpatterns = [
     path("", views.index, name="index"),
     path("admin/", admin.site.urls),
+    path('api-token-auth/', views.UserObtainTokenView.as_view()),
```
rmoore-theta, line 38: Was it intentional to use this url instead of conforming to the others by starting with api/?

jwhite3:
To do: 
- [x] Expire user token

blueprintapi/tests.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework.test import APITestCase
+
+
+class UnauthenticatedAPITestCase(APITestCase):
+    def test_unauthenticated_request_returns_403(self):
+        test_cases = ('/api/projects/', '/api/users/')
```
jwhite3, line 10: This list certainly isn't exhaustive, but it's really only meant to test that without a token, you get an unauthenticated error. Some endpoints don't have these restrictions though.

blueprintapi/permissions.py
```diff
@@ -0,0 +1,13 @@
+from rest_framework.permissions import DjangoObjectPermissions
+
+
+class StrictDjangoObjectPermissions(DjangoObjectPermissions):
```
jwhite3, line 4: Forgot a Docstring here. I'll add it

blueprintapi/urls.py
jwhite3, line 38: I intentionally didn't put the auth token endpoint under the `/api` path, but if folks would prefer that it gets moved there, let me know.

blueprintapi/views.py
```diff
@@ -17,3 +20,17 @@ def healthcheck(request):
             raise OperationalError("Could not connect to database")
     response = {"content": "Healthy"}
     return Response(response)
+
+
+class UserObtainTokenView(ObtainAuthToken):
```
jwhite3, line 25: Need to extend this to expire the token.

catalogs/views.py
```diff
@@ -1,81 +1,49 @@
...
 
 class CatalogListView(generics.ListCreateAPIView):
-    """
-    Use for read-write endpoints to represent a collection of model instances.
+    """Use for read-write endpoints to represent a collection of model instances.
     Provides get and post method handlers.
     """
-
     queryset = Catalog.objects.all().order_by("pk")
+    permission_classes = [AllowAny, ]
```
jwhite3, line 18: `AllowAny` is open permissions, but let me know if this should be read-only.

catalogs/views.py
```diff
@@ -1,81 +1,49 @@
...
     serializer_class = CatalogListSerializer
 
-    def list(self, request):
-        queryset = self.get_queryset()
-        serializer = CatalogListSerializer(queryset, many=True)
-        return Response(serializer.data, status=status.HTTP_200_OK)
 
+class CatalogControlBaseView(generics.GenericAPIView):
+    queryset = Catalog.objects.all()
+    permission_classes = [AllowAny, ]
```
jwhite3, line 24: Same with controls, should this be read-only?

components/tests.py
```diff
@@ -176,16 +173,18 @@ def setUpTestData(cls):
...
             description="This one is better.",
-            catalog=Catalog.objects.get(id=cls.test_catalog.id),
+            catalog=Catalog.objects.get(id=test_catalog.id),
             controls=TEST_COMPONENT_CONTROLS,
             search_terms=["cool", "best"],
             type="software",
             component_json=TEST_COMPONENT_JSON_BLOB,
         )
 
+        cls.test_catalog = test_catalog
```
jwhite3: Didn't mean to add this in this PR. I'll remove.

components/views.py
```diff
@@ -1,67 +1,63 @@
...
     """
-
     queryset = Component.objects.all().order_by("pk")
     serializer_class = ComponentListSerializer
-
-    def list(self, request):
-        queryset = self.get_queryset()
-        serializer = ComponentListSerializer(queryset, many=True)
-        return Response(serializer.data, status=status.HTTP_200_OK)
+    filter_backends = [ObjectPermissionsFilter, ]
```
jwhite3, line 19: For list views, we need to be sure to add this permissions filter: only allowed objects should be returned for list views.

components/views.py
```diff
@@ -1,67 +1,63 @@
...
-        type_list = Component.objects.order_by().values_list("type").distinct()
-        return Response(type_list, status=status.HTTP_200_OK)
+class ComponentTypeListView(generics.ListAPIView):
+    queryset = Component.objects.order_by().values_list("type").distinct()
+    filter_backends = [ObjectPermissionsFilter, ]
+
+    def list(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset())
+
+        return Response(queryset, status=status.HTTP_200_OK)
```
jwhite3, line 63: This isn't ideal, and we should probably revisit this at some point. Is the purpose just to get a set of available component types?

projects/tests.py
```diff
@@ -321,15 +320,12 @@ def setUpTestData(cls):
...
-        response = client.get(
+        response = self.client.get(
             reverse("project-detail", kwargs={"project_id": self.test_project.pk})
         )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
 
         project = self.test_project
-        serializer = ProjectSerializer(project)
-
-        self.assertEqual(response.data, serializer.data)
```
jwhite3, line 331: This was interesting, and I'm actually not quite sure what was going on. As far as I could tell, the contents were the same, but testing for equality was failing due to some of the types in each of these `data` attributes. 

Need to investigate further, but since we were checking content additionally below, I removed this for now.

projects/tests.py
```diff
@@ -394,11 +390,16 @@ def test_invalid_project(self):
         self.assertEqual(resp.status_code, 404)
 
     def test_invalid_project_permissions(self):
+        user, _ = User.objects.get_or_create(username='invalid_perms')
+        token, _ = Token.objects.get_or_create(user=user)
```
jwhite3, line 394: This temporary user has no permissions assigned.

projects/tests.py
```diff
@@ -394,11 +390,16 @@ def test_invalid_project(self):
...
+
+        self.client.force_authenticate(user=user, token=token)
```
jwhite3, line 396: This is how you can either authenticate as a different user for a particular test. If you want to remove authentication, you can set `user` and `token` to `None` to ensure the request is submitted without any credentials.

projects/views.py
```diff
@@ -2,83 +2,35 @@
...
-            )
-        project_instance.delete()
-        return Response(
-            {"response": "You have succesfully deleted the project!"},
-            status=status.HTTP_200_OK,
-        )
 
+class ProjectAddComponentView(generics.GenericAPIView):
+    queryset = Project.objects.all()
+    lookup_field = "pk"
```
jwhite3: Need to actually use this in the `post` method. I forgot to add it in.

projects/views.py
```diff
@@ -87,50 +39,58 @@ def post(self, request, *args, **kwargs):
...
-            {"response": "Incompatable catalog selected"},
+            {"response": "Incompatible catalog selected"},
             status=status.HTTP_400_BAD_REQUEST,
         )
 
 
-class ProjectRemoveComponentView(APIView):
+class ProjectRemoveComponentView(generics.GenericAPIView):
+    queryset = Project.objects.prefetch_related('components')
+
```
jwhite3, line 50: Need to add and utilize `lookup_field` or `lookup_url_kwarg` here.

projects/views.py
```diff
@@ -148,26 +108,31 @@ def get(self, request, project_id, *args, **kwargs):
...
-        response.append({"project": project_serializer.data})
-        response.append({"components": serializer.data})
-        response.append({"total_item_count": paginator.count})
-        response.append({"type_list": type_list})
+        response = [
+            {"project": project_serializer.data},
+            {"components": serializer.data},
+            {"total_item_count": paginator.count},
+            {"type_list": type_list}
+        ]
```
jwhite3: This needs a serializer so that we can utilize `generics` more, but that's out of scope of this PR.

projects/views.py
```diff
@@ -148,26 +108,31 @@ def get(self, request, project_id, *args, **kwargs):
...
 
-class ProjectComponentNotAddedListView(APIView):
-    def get(self, request, project_id, *args, **kwargs):
-        project_instance = get_object_or_404(Project, pk=project_id)
+class ProjectComponentNotAddedListView(generics.GenericAPIView):
+    queryset = Project.objects.all()
+    lookup_url_kwarg = "project_id"
+    serializer_class = ComponentListBasicSerializer
+
+    def get(self, request, *args, **kwargs):
```
jwhite3, line 127: Need to overwrite `list` instead of `get`. I'll update.

pyproject.toml
```diff
@@ -19,6 +19,7 @@ pathlib = "^1.0.1"
 jsonschema = "^4.6.0"
 django-filter = "^22.1"
 pydantic = "^1.9.1"
+requests = "^2.28.1"
```
jwhite3, line 22: This was being used in `Environment.py`, but was not a project requirement.

users/models.py
```diff
@@ -1,6 +1,6 @@
 # from django.db import models
-from django.contrib.auth.models import AbstractUser
+from django.contrib.auth.models import AbstractUser, PermissionsMixin
 
 
-class User(AbstractUser):
-    pass
+class User(AbstractUser, PermissionsMixin):
+    ...
```
jwhite3: I go back and forth on this one (`...` vs `pass`). We should probably make a distinction for the project about when to use one vs the other. I think I'll put `pass` back for now.

components/tests.py
jwhite3: Actually, I realized that that class attribute was never used, so I'm going to go ahead and update.

projects/views.py
jwhite3: Nevermind. I think this would require a more significant refactor, so I'll hold on this one.

blueprintapi/urls.py
rmoore-theta: I only suggested this for consistency with the other endpoints that the UI calls, but the front end related code already is handling this exception to the rule, so it is fine with me.

projects/views.py
rmoore-theta: Would love to talk offline about this so I can understand more :)

users/models.py
rmoore-theta: This could be a good dev brainstorming discussion to understand the differences :)

projects/views.py
jwhite3: Just kidding, this would require updates to the url to utilize. Need to address in a later PR.

projects/views.py
jwhite3: ~I lied, I fixed it.~

jwhite3: Nope, still needs more of a refactor. Excluding this for this PR.

projects/views.py
jwhite3: This would need an update to the url patterns, so skipping for this PR.

blueprintapi/settings.py
```diff
@@ -122,6 +121,7 @@
 }
 
 AUTH_USER_MODEL = "users.User"
+AUTH_TOKEN_TTL = 24  # Hours
```
jwhite3, line 124: We should figure out what the default should be. Also, this would be great to have as an environment variable, but I want to wait to deal with that until we remove/refactor `Environment.py` and use the `django-environ` plugin.

blueprintapi/authentication.py
```diff
@@ -0,0 +1,30 @@
...
+        if not token.user.is_active:
+            raise AuthenticationFailed(_('User inactive or deleted.'))
+
+        # This is required for the time comparison
+        utc_now = datetime.utcnow()
+        utc_now = utc_now.replace(tzinfo=pytz.utc)
+
+        if utc_now - token.created > timedelta(hours=settings.AUTH_TOKEN_TTL):
+            token.delete()
+            raise AuthenticationFailed(_('Token has expired.'))
```
jwhite3, line 28: @amyd11 This will delete the user's token if it's "expired" and result in a 401 "Token has expired" response. The front end will have to redirect to the login page, and then once the user logs in again, a new token will be created.

blueprintapi/settings.py
jauschalley: Praise: Thanks for doing this.  I believe the default should be 4 hours but I'll get confirmation on that.  It's definitely good to be merged as is for the time being.  Of course, validating that the expiration is working is more difficult the longer that the token lives.

pyproject.toml
jauschalley: Praise: Great catch!

catalogs/views.py
jauschalley: Question:  Is this the temporary work around that we discussed until we follow up with more exacting permissions?

blueprintapi/urls.py
jauschalley: I doubt this is going to hurt anyone's feelings.

jauschalley (dismissed):
Praise:  Thanks for taking on this task being so thorough in your work.  I'm approving with the gentle reminder that this is on hold until the front end work is complete.  Thanks!

catalogs/views.py
jwhite3: Exactly. I wanted to point it out here just in case someone did have a thought before we start with this.

blueprintapi/settings.py
jwhite3: Awesome, let me know. Super easy to change and since we have it as a settings value (which I'm pretty sure makes it configurable for tests) it should be very easy to write a test for it. I just forgot lol. Let me see about adding a test for this. 

I did confirm that it works "manually" since I already had tokens older than that value.

jwhite3: All good: https://github.com/CMSgov/blueprint_api/pull/66/commits/8d0ff09d0cd2a4e72e36b57372a51274b0ae0a1f

jauschalley (approved):

projects/views.py
```diff
@@ -2,83 +2,34 @@
...
-        try:
-            return Project.objects.get(id=project_id)
-        except Project.DoesNotExist:
-            return None
 
-    def get(self, request, project_id, *args, **kwargs):
-        project_instance = get_object_or_404(Project, pk=project_id)
-        serializer = ProjectSerializer(project_instance)
-        return Response(serializer.data, status=status.HTTP_200_OK)
+class ProjectsDetailView(generics.RetrieveUpdateDestroyAPIView):
```
amyd11, line 21: Love that using the generics here cleans up the code and brings consistency in implementation!

amyd11 (approved):

PR #68, opened 2022-08-04
added next control id to response for navigating to the next one

rmoore-theta:
*What does this PR do?*
Adds the next control id 
*Jira ticket number?*
[ISPGBSS-1179](https://jiraent.cms.gov/browse/ISPGBSS-1179)

*Changes*

- adds the next control id to the response

*Testing*

- should now be able to see the next control id when getting a control

### DIFF
--- a/catalogs/catalogio.py
+++ b/catalogs/catalogio.py
@@ -120,6 +120,16 @@ class CatalogTools(object):
         )
         return result_dict
 
+    def get_next_control_by_id(self, control_id: str):
+        search_collection = self.get_control_ids()
+        next = False
+        for item in search_collection:
+            if next is True:
+                return item
+            if item == control_id:
+                next = True
+        return ""
+
     def get_control_statement(self, control: dict) -> List:
         statement = self.get_control_part_by_name(control, "statement")
         text: List[dict] = []
@@ -219,6 +229,7 @@ class CatalogTools(object):
             "description": self.__get_simplified_prose(desc),
             "implementation": implementation.get("prose") if implementation else "",
             "guidance": guidance.get("prose") if guidance else "",
+            "next_id": self.get_next_control_by_id(control_id),
         }
 
         return control_data
--- a/catalogs/catalogio.py
+++ b/catalogs/catalogio.py
@@ -88,7 +88,7 @@ class CatalogTools(object):
 
     def get_control_ids(self) -> List:
         search_collection = self.get_controls()
-        return [item.get("id") for item in search_collection]
+        return sorted(item.get("id") for item in search_collection)
 
     def get_controls_all(self) -> List:
         controls: List[dict] = []
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+---
+version: 2
+updates:
+    - package-ecosystem: github-actions
+      directory: /
+      schedule:
+          interval: daily
+
+    - package-ecosystem: pip
+      directory: /
+      schedule:
+          interval: weekly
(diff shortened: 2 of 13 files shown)


jauschalley (dismissed):

catalogs/catalogio.py
```diff
@@ -120,6 +120,16 @@ def get_control_by_id(self, control_id: str) -> dict:
...
 
+    def get_next_control_by_id(self, control_id: str):
+        search_collection = self.get_control_ids()
+        next = False
+        for item in search_collection:
+            if next is True:
+                return item
+            if item == control_id:
+                next = True
+        return ""
```
jwhite3, line 131: What if `control_id` isn't in the list of ids? That should probably raise an error. here's an example that handles that bit explicitly:
```python
try:
   next_idx = search_collection.index(control_id) + 1
except ValueError as e:  # Provided control_id isn't in the list
    raise ValueError('Cannot determine next control. Provided id does not exist') from e

try:
    return search_collection[next_idx]  # control_id was at the end of the list
except IndexError:
    return ""
```

But also, are we guaranteeing an order somewhere such that this will return the same "next" id each time for the same set of control ids?

rmoore-theta: I did not want to throw an exception, because when you reach the end of the list, it will not find another control and just return an empty string. The UI would then be sent back to the projects controls page where they would see what controls they have to still work on. 

Is it possible to only consider it an exception if not at the end of the list of controls?

The order is determined by the get_control_ids, I am not sure if we need to guarantee these will be in a specific order in this function :/ but I am not sure what would cause the order to change for a given list of controls.

jwhite3: > I did not want to throw an exception, because when you reach the end of the list, it will not find another control and just return an empty string.

This sounds fine for the case where there is no "next" control because the input id was at the end of the list

> Is it possible to only consider it an exception if not at the end of the list of controls?

The exception in my example is raised when the provided control id doesn't exist in the list regardless of position. I think that this makes sense because returning an empty string for both a non-existent control and a terminal control is ambiguous. 

> I am not sure if we need to guarantee these will be in a specific order in this function :/ 

Then how do you define "next" if there's no guaranteed order to the list of control ids?

rmoore-theta: > Then how do you define "next" if there's no guaranteed order to the list of control ids?
I was defining next as the next item found. Do you have a suggestion on how to improve this?

jwhite3: I guess that's my point: defining "next" as the next item found in a list with no guaranteed order means that you might get different results with each request (including duplications).

The solution would be to make sure that the controls are sorted, so that you get the same result each time. You could sort by whatever you want or whatever is most convenient (id, name, etc)

rmoore-theta: So looking at where the list of controls comes from, it is a list of control_ids from get_control_ids > get_controls gets a list of the groups to iterate over from get_groups > the groups come out of the oscal, which is a json file. 

It is my understanding that the oscal file is validated when initially uploaded, which should prevent the duplications.

As far as I am aware, we do not have anything that would update the json file to order it differently.

jwhite3: The duplications wouldn't result from there being duplications in the data file itself, but rather there's a potential that if subsequent requests result in a different "next" you may get the same value as in a previous request unless you're guaranteed that the order is the same.

jwhite3:
@rmoore-theta would you mind updating this branch?

jauschalley (approved):

jauschalley:
Chore: Let's just close this PR, revert back to the commit before the rebase, and try a fresh rebase.  We need to keep our commit history clean.

jauschalley (changes requested):
Chore: Let's just close this PR, revert back to the commit before the rebase, and try a fresh rebase. We need to keep our commit history clean.

PR #69, opened 2022-08-04
Add unittests Actions workflow

jwhite3:
*What does this PR do?*
This PR adds a new GitHub Actions workflow that will execute unit tests on push to main branches `main`, `develop` as well as all pull requests.

*Jira ticket number?*
https://jiraent.cms.gov/browse/ISPGBSS-1137

*Changes*
- Add test workflow

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+---
+version: 2
+updates:
+    - package-ecosystem: github-actions
+      directory: /
+      schedule:
+          interval: daily
+
+    - package-ecosystem: pip
+      directory: /
+      schedule:
+          interval: weekly
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,42 @@
+---
+name: Test
+on:
+  push:
+    branches:
+      - main
+      - develop
+  pull_request:
+
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_DB: postgres
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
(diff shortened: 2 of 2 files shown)


jwhite3:
@jauschalley Mind adding the corresponding Jira info to this PR? Still don't have access :(

.github/workflows/test.yaml
```diff
@@ -0,0 +1,42 @@
+---
+name: Test
+on:
+  push:
+    branches:
+      - main
+      - develop
+  pull_request:
+
```
jauschalley, line 9: Do this mean that it will run on every pull request regardless of branch?

jauschalley:
Are you seeing the 'actions' tab in your view?  If not, is this PR blocked by not being able to complete the setup?

.github/workflows/test.yaml
jwhite3: Yep. In the past I've found this pretty useful and we can always update if we decide that it's not helpful on this project.

jauschalley: That sounds good to me.

jwhite3:
> Are you seeing the 'actions' tab in your view? If not, is this PR blocked by not being able to complete the setup?

No Actions tab, but I don't think that necessarily has to be a blocker. We just won't know if there's a bug or something until we've enabled actions.

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+updates:
+    - package-ecosystem: github-actions
+      directory: /
+      schedule:
+          interval: daily
+
+    - package-ecosystem: pip
+      directory: /
+      schedule:
+          interval: weekly
```
jwhite3, line 12: Figured this might also be something that we'd want to include. Let me know if the preference would be to add this in a separate PR and I'll pop it out and add it separately.

jauschalley: Seems Fine ™️  to stay in this PR

jauschalley (approved):

rmoore-theta (approved):

PR #74, opened 2022-08-08
Fix syntax error in test workflow

jwhite3:
*What does this PR do?*
Fixes a syntax error in `.github/workflows/test.yaml`.

*Jira ticket number?*
N/A

*Changes*
- Remove extra "-" in `test.yaml`

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -37,6 +37,6 @@ jobs:
         run: pip install poetry
       - run: poetry config virtualenvs.create false
       - name: Install dependencies
-      - run: poetry install --no-dev --no-ansi --no-interaction
+        run: poetry install --no-dev --no-ansi --no-interaction
       - name: Run tests
         run: python3 manage.py test


jwhite3:
Good news! Fixed the syntax error, but it also looks like I didn't correctly include the environment variables when running the test :(

.github/workflows/test.yaml
```diff
@@ -37,6 +37,6 @@ jobs:
         run: pip install poetry
       - run: poetry config virtualenvs.create false
```
rmoore-theta, line 38: Does this line need to remove the - as well?

jwhite3: Nope! I just didn't give a particular name for that step.

jwhite3: The error was that there was an orphaned `- name` because of the extra "-" in front of the following `run` section.

rmoore-theta (approved):

jauschalley (approved):

PR #78, opened 2022-08-10
added a sort for get_control_ids

rmoore-theta:
*What does this PR do?*
Adds the next control id
*Jira ticket number?*
[ISPGBSS-1179](https://jiraent.cms.gov/browse/ISPGBSS-1179)

*Changes*

- adds the next control id to the response

*Testing*

- should now be able to see the next control id when getting a control

### DIFF
--- a/catalogs/catalogio.py
+++ b/catalogs/catalogio.py
@@ -124,12 +124,12 @@ class CatalogTools(object):
         search_collection = self.get_control_ids()
         try:
             next_idx = search_collection.index(control_id) + 1
-            return search_collection[next_idx]  # control_id was at the end of the list
+            return search_collection[next_idx]
         except ValueError as e:
             raise MissingControlError(
                 "Cannot determine next control. Provided id does not exist"
             ) from e
-        except IndexError:
+        except IndexError:  # control_id was at the end of the list
             return ""
 
     def get_control_statement(self, control: dict) -> List:
--- a/catalogs/catalogio.py
+++ b/catalogs/catalogio.py
@@ -122,13 +122,15 @@ class CatalogTools(object):
 
     def get_next_control_by_id(self, control_id: str):
         search_collection = self.get_control_ids()
-        next = False
-        for item in search_collection:
-            if next is True:
-                return item
-            if item == control_id:
-                next = True
-        return ""
+        try:
+            next_idx = search_collection.index(control_id) + 1
+            return search_collection[next_idx]  # control_id was at the end of the list
+        except ValueError as e:
+            raise MissingControlError(
+                "Cannot determine next control. Provided id does not exist"
+            ) from e
+        except IndexError:
+            return ""
 
     def get_control_statement(self, control: dict) -> List:
         statement = self.get_control_part_by_name(control, "statement")
@@ -245,3 +247,7 @@ class CatalogTools(object):
     def catalog_title(self) -> str:
         metadata = self.oscal.get("metadata", {})
         return metadata.get("title", "")
+
+
+class MissingControlError(ValueError):
+    pass
--- a/catalogs/catalogio.py
+++ b/catalogs/catalogio.py
@@ -120,6 +120,16 @@ class CatalogTools(object):
         )
         return result_dict
 
+    def get_next_control_by_id(self, control_id: str):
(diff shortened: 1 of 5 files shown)


jauschalley (dismissed):

catalogs/catalogio.py
```diff
@@ -120,6 +120,16 @@ def get_control_by_id(self, control_id: str) -> dict:
...
 
+    def get_next_control_by_id(self, control_id: str):
+        search_collection = self.get_control_ids()
+        next = False
+        for item in search_collection:
+            if next is True:
+                return item
+            if item == control_id:
+                next = True
+        return ""
```
jwhite3: I still think an error should be raised if the `control_id` isn't found in the `search_collection` list. I don't think there's anything at the view layer that checks this either, so there'd be no distinction between the end of the list and a non-existent control.

This would require a couple of things: 
- Raise an error (possibly a custom error so it's easier to handle downstream)
- Handle the error in the appropriate view(s) and return a 404 response with a clear message.
- Additional test for trying to access a non-existent control

jwhite3 (dismissed):

catalogs/catalogio.py
rmoore-theta: This should not raise an error, it should just end with an empty string being returned. If that is not what will happen, then that is unexpected.

rmoore-theta: Just for a little more context; 
If you are on the last control, then this is the information to display the page for the last control. Which is why I say this should not be an error, cause then you could not complete the last control.

jwhite3: > This should not raise an error, it should just end with an empty string being returned. If that is not what will happen, then that is unexpected.

I'm saying that an error _should_ be raised, not that there actually is one currently. I'm saying that the current ambiguity between the "next" control from a non-existent input `control_id` and the input `control_id` already being the last in the list needs to be addressed and handled.

catalogs/catalogio.py
```diff
@@ -120,6 +120,18 @@ def get_control_by_id(self, control_id: str) -> dict:
         )
         return result_dict
 
+    def get_next_control_by_id(self, control_id: str):
+        search_collection = self.get_control_ids()
+        try:
+            next_idx = search_collection.index(control_id) + 1
+            return search_collection[next_idx]  # control_id was at the end of the list
```
jwhite3: This comment is a good one, but I put it in the wrong spot. Should be down next to the `IndexError`. Not a blocker though

components/tests.py
```diff
@@ -37,27 +37,27 @@
...
                             {
                                 "uuid": "73dd3c2e-54eb-43c6-a488-dfb7c79d9413",
-                                "control-id": "ac-2.2",
+                                "control-id": "ac-2",
                                 "description": "This component statisfies b.",
                             },
                             {
                                 "uuid": "73dd3c2e-54eb-43c6-a488-dfb7c79d9413",
-                                "control-id": "ac-2.3",
+                                "control-id": "at-1",
```
jwhite3, line 50: Remind me why the control id needed to change? I think you mentioned this, but I forget.

components/tests.py
```diff
@@ -218,7 +218,7 @@ def setUpTestData(cls):
             title="Cool Component",
             description="Probably the coolest component you ever did see. It's magical.",
             catalog=Catalog.objects.get(id=cls.test_catalog.id),
-            controls=["ac-2.1", "ac-6.10", "ac-8", "au-6.1", "sc-2"],
+            controls=["ac-1", "ac-2", "at-1", "at-2", "at-3", "au-1", "au-2"],
```
jwhite3, line 221: Was this because these controls didn't exist?

components/tests.py
rmoore-theta: For the catalog and components controls to be aligned. Nothing checked this before

components/tests.py
rmoore-theta: Correct

catalogs/catalogio.py
rmoore-theta: Updated

jwhite3 (approved):

jauschalley (approved):

PR #79, opened 2022-08-10
Bug/open create user perms

jwhite3:
*What does this PR do?*
This PR updates user permissions to allow for new user creation. In addition, the health check endpoint was update with read only permissions.

*Jira ticket number?*
N/A

*Changes*
- Fix `UserSerializer` to correctly receive/apply required user data
- Add permission class to health check view
- Add post_save signal to assign user permissions to the user on creation
- Add basic project permissions for new user (so that they can create a new project and whatnot)
- Fix creator requirement in `ProjectListSerializer`

Needs coordination with:
 - [ ] https://github.com/CMSgov/blueprint_ui/pull/71

### DIFF
--- a/projects/tests.py
+++ b/projects/tests.py
@@ -399,7 +399,7 @@ class ProjectAddComponentViewTest(AuthenticatedAPITestCase):
             "/api/projects/add-component/",
             {"creator": 0, "component_id": 1, "project_id": self.test_project.id},
         )
-        self.assertEqual(resp.status_code, 403)
+        self.assertEqual(resp.status_code, 404)
 
     def test_invalid_component(self):
         resp = self.client.post(
--- a/projects/tests.py
+++ b/projects/tests.py
@@ -13,8 +13,6 @@ from projects.models import Project
 from testing_utils import AuthenticatedAPITestCase
 from users.models import User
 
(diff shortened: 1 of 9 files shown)


jwhite3:
To do: 
- [x] Add/update tests

jauschalley (dismissed):
LGTM

projects/tests.py
```diff
@@ -399,7 +399,7 @@ def test_invalid_project_permissions(self):
             "/api/projects/add-component/",
             {"creator": 0, "component_id": 1, "project_id": self.test_project.id},
         )
-        self.assertEqual(resp.status_code, 403)
+        self.assertEqual(resp.status_code, 404)
```
jwhite3, line 395: This update is needed because before, users had 0 default permissions, but now they have some model-level permissions. The 403 was a result of model-level permissions checks being thrown, while the 404 is due to this user not having object-level permissions for the particular project being tested.

jwhite3:
@jauschalley @amyd11 I think this is ready to go. Kept finding some holes that needed to be plugged, but this should allow for basic usage until we can iron out some other stuff.

jauschalley (approved):

amyd11 (approved):

PR #83, opened 2022-08-12
1166 - Adding Project-specific Controls.

TMCamp:
*What does this PR do?*
This PR adds a Control model to the Catalog app and when a Catalog is added, writes all of the Controls in the Catalog to the Control table. 

It also adds a Project Controls model to the Projects app. When  a Project is created the Controls associated with the Project's Catalog (from the catalogs.Controls) table are associated with the Project through the ProjectControls model. This allows us to update the Control status per project.

*Jira ticket number?*

*Changes*

- Added Controls model to catalogs
- Added hook to write Catalog controls to the Controls table upon Catalog creation
- Added ProjectControls model to projects
- Added hook to associate Controls with a Project upon Project creation

*Testing*

- [ ] Step 1
- [ ] Step 2


catalogs/catalogio.py
```diff
@@ -12,15 +12,13 @@ def __init__(self, source, text=False):
...
-            self.info["groups"] = self.get_groups()
+            self.info = {"groups": self.get_groups()}
         except Exception:
             self.oscal = None
             self.status = "error"
             self.status_message = "Error loading catalog"
             self.catalog_id = None
-            self.info = {}
-            self.info["groups"] = None
+            self.info = {"groups": None}
```
jwhite3, line 21: This is out of scope for this PR, but I wanted to mention it so that we could address this in the future:

What was the reasoning for setting these attributes if there's an error? This actually can cause a bunch of odd downstream errors that are difficult to debug. For example, with `self.oscal` being set to `None`, if `catalog_title` is called, you'll get a mysterious attribute error rather than something that says that there was an issue reading the catalog. I think it would be better to go ahead and raise an informative error if there's a problem loading a catalog.

catalogs/models.py
```diff
@@ -33,17 +42,47 @@ class Catalog(models.Model):
...
     created = models.DateTimeField(auto_now_add=True, db_index=True)
     updated = models.DateTimeField(auto_now=True, db_index=True, null=True)
 
     def __str__(self):
         return self.name
 
 
+@receiver(post_save, sender=Catalog)
+def add_controls(sender, instance, **kwargs):
+    if kwargs["created"]:
```
jwhite3, line 66: Not a blocker or anything, but `created` is actually available to use here:
```suggestion
def add_controls(sender, instance, created, **kwargs):
    if created:
```

catalogs/models.py
```diff
@@ -33,17 +42,47 @@ class Catalog(models.Model):
...
+            for c in controls:
+                control_data = catalog.get_control_data_simplified(c)
+                ctrl = Controls(
+                    catalog=instance,
+                    control_id=c,
+                    control_label=control_data.get("label"),
+                    sort_id=control_data.get("sort_id"),
+                    title=control_data.get("title"),
+                )
+                ctrl.save()
```
jwhite3, line 79: This would be a good candidate for a [bulk create](https://docs.djangoproject.com/en/4.1/ref/models/querysets/#bulk-create)

projects/models.py
```diff
@@ -101,64 +105,97 @@ def __str__(self):
...
-    except ObjectDoesNotExist as e:
-        logger.warning(f"Inherited components not found: {e}")
+        try:
+            instance.components.add(default)
+        except Exception as e:
+            logger.error(f"Could not create default component: {e}")
 
 
 @receiver(post_save, sender=Project)
-def post_create_setup(sender, instance, created, **kwargs):
```
jwhite3, line 146: Looks like something funky might've happened with a rebase here

jwhite3:
@TMCamp This is looking great! I like this direction

catalogs/catalogio.py
TMCamp: Yeah, agree. The initial code came from the compliance-io library and I can't recall why it was done that way.

PR #84, opened 2022-08-12
Feature/1166 project control

TMCamp:
*What does this PR do?*
This PR adds a Control model to the Catalog app and when a Catalog is added, writes all of the Controls in the Catalog to the Control table. 

It also adds a Project Controls model to the Projects app. When  a Project is created the Controls associated with the Project's Catalog (from the catalogs.Controls) table are associated with the Project through the ProjectControls model. This allows us to update the Control status per project.

*Jira ticket number?*

*Changes*

- Added Controls model to catalogs
- Added hook to write Catalog controls to the Controls table upon Catalog creation
- Added ProjectControls model to projects
- Added hook to associate Controls with a Project upon Project creation

*Testing*

- [ ] Step 1
- [ ] Step 2


catalogs/catalogio.py
```diff
@@ -12,15 +12,13 @@ def __init__(self, source, text=False):
             self.status = "ok"
             self.status_message = "Success loading catalog"
             self.catalog_id = self.oscal.get("id")
-            self.info = {}
-            self.info["groups"] = self.get_groups()
-        except Exception:
+            self.info = {"groups": self.get_groups()}
+        except CatalogLoadError:
```
jwhite3: I like the new error that makes it obvious where this went wrong, but this will actually never happen since `CatalogLoadError` won't be raised by anything in that try block.

Instead you'd need to catch the exceptions that may be raised like so:
```python
    try:
        ...
    except (IOError, FileNotFoundError, json.decoder.JSONDecodeError) as e:
        raise CatalogLoadError(f'oops.. can't load catalog {source}') from e  # Chain the exceptions
```

Of course the message needs to be more useful and whatnot, but this is the basic idea.

catalogs/catalogio.py
```diff
@@ -12,15 +12,13 @@ def __init__(self, source, text=False):
...
             self.oscal = None
             self.status = "error"
             self.status_message = "Error loading catalog"
             self.catalog_id = None
-            self.info = {}
-            self.info["groups"] = None
+            self.info = {"groups": None}
```
jwhite3: 👍

catalogs/models.py
```diff
@@ -33,17 +42,47 @@ class Catalog(models.Model):
...
+            for c in controls:
+                control_data = catalog.get_control_data_simplified(c)
+                ctrl = Controls(
+                    catalog=instance,
+                    control_id=c,
+                    control_label=control_data.get("label"),
+                    sort_id=control_data.get("sort_id"),
+                    title=control_data.get("title"),
+                )
+                ctrl.save()
```
jwhite3: I mentioned this in the other PR, but I think this may be a good candidate for [bulk_create](https://docs.djangoproject.com/en/4.1/ref/models/querysets/#bulk-create)

catalogs/models.py
```diff
@@ -67,3 +106,30 @@ def auto_delete_file_on_change(sender, instance, **kwargs):
     if not old_file == new_file:
         if os.path.isfile(old_file.path):
             os.remove(old_file.path)
+
+
+class Controls(models.Model):
+    catalog = models.ForeignKey(to="catalogs.Catalog", on_delete=models.CASCADE)
```
jwhite3, line 116: Just curious, why add the `to` keyword?

catalogs/models.py
```diff
@@ -33,17 +42,47 @@ class Catalog(models.Model):
...
     created = models.DateTimeField(auto_now_add=True, db_index=True)
     updated = models.DateTimeField(auto_now=True, db_index=True, null=True)
 
     def __str__(self):
         return self.name
 
 
+@receiver(post_save, sender=Catalog)
+def add_controls(sender, instance, **kwargs):
+    if kwargs["created"]:
```
jwhite3: Not a blocker, just a preference:
```suggestion
def add_controls(sender, instance, created, **kwargs):
    if created:
```

catalogs/models.py
```diff
@@ -67,3 +106,30 @@ def auto_delete_file_on_change(sender, instance, **kwargs):
     if not old_file == new_file:
         if os.path.isfile(old_file.path):
             os.remove(old_file.path)
+
+
+class Controls(models.Model):
+    catalog = models.ForeignKey(to="catalogs.Catalog", on_delete=models.CASCADE)
+    control_id = models.CharField(
+        max_length=12,
+        unique=False,
```
jwhite3: I think `unique=False` is the default behavior, so no need to specify unless `unique=True`.

catalogs/models.py
```diff
@@ -67,3 +106,30 @@ def auto_delete_file_on_change(sender, instance, **kwargs):
...
+        blank=False,
```
jwhite3: Same for `blank` actually

projects/models.py
```diff
@@ -1,88 +1,92 @@
...
-    IMPACT_LEVEL_CHOICES = [
-        ("low", "Low"),
-        ("moderate", "Moderate"),
-        ("high", "High"),
-        ("pii/phi", "PII or PHI"),
-    ]
+    controls = models.ManyToManyField(
+        to="catalogs.Controls",
+        through="ProjectControl",
+        blank=True,
```
jwhite3: I'm not sure `blank` makes sense here. Should be fine to exclude that option. If a project doesn't have any controls associated with it, the manager will just return an empty QuerySet

projects/models.py
```diff
@@ -101,14 +105,30 @@ def __str__(self):
         return self.title
 
 
+def add_project_controls(instance: Project):
+    try:
+        controls = apps.get_model("catalogs", "Controls")
+        catalog_controls = controls.objects.filter(catalog_id=instance.catalog)
+        print(catalog_controls)
```
jwhite3: I assume this was for debugging, but we should take it out before merging. Or switch it to a log

projects/models.py
```diff
@@ -101,14 +105,30 @@ def __str__(self):
...
+        if catalog_controls:
```
jwhite3: We should consider using [exists](https://docs.djangoproject.com/en/4.1/ref/models/querysets/#exists) here:
```suggestion
        if catalog_controls.exists():
```

projects/models.py
```diff
@@ -101,14 +105,30 @@ def __str__(self):
...
+            instance.controls.set(
+                catalog_controls,
+                through_defaults={"status": ProjectControl.Status.NOT_STARTED},
+            )
+    except ObjectDoesNotExist as e:
```
jwhite3: I'm not sure how this exception will be raised in this case. Controls are filtered, and then before attempting to set controls, the QuerySet is checked to make sure that it's not empty. Since the input controls to `set` are coming from the DB itself, I'm not sure how the objects won't exist  in the DB.

projects/models.py
```diff
@@ -162,3 +187,26 @@ def post_create_setup(sender, instance, created, **kwargs):
...
+
+
+class ProjectControl(models.Model):
+    class Status(models.TextChoices):
+        NOT_STARTED = "not_started", _("Not started")
+        INCOMPLETE = "incomplete", _("Incomplete")
+        COMPLETE = "completed", _("Completed")
+
+    project = models.ForeignKey(to="projects.Project", on_delete=models.CASCADE)
+    controls = models.ForeignKey(to="catalogs.Controls", on_delete=models.PROTECT)
```
jwhite3: Not a blocker, and definitely nitpicky:

Shouldn't "controls" be singular in this context?

jwhite3: Nice, good idea to use `PROTECT` in this case

jwhite3 (commented):
This is looking great. Thanks for all the work in adding the through table

projects/models.py
TMCamp: Yep

catalogs/catalogio.py
TMCamp: Added _slightly_ more useful message and logging.

catalogs/models.py
TMCamp: Yep, sounds good.

catalogs/models.py
TMCamp: I added the `to=` keywords while I was struggling with some logic and it just helped to keep it clear.

blueprintapi/settings.py
```diff
@@ -187,11 +187,13 @@
...
+        "blueprintapi.authentication.ExpiringTokenAuthentication",
     ],
-    'DEFAULT_PERMISSION_CLASSES': [
-        'blueprintapi.permissions.StrictDjangoObjectPermissions',
+    "DEFAULT_PERMISSION_CLASSES": [
+        "blueprintapi.permissions.StrictDjangoObjectPermissions",
     ],
     "DEFAULT_SCHEMA_CLASS": "rest_framework.schemas.coreapi.AutoSchema",
+    "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.PageNumberPagination",
+    "PAGE_SIZE": 20,
```
jwhite3: We probably don't want to do this yet. Instead, you should be able to specify the pagination class as an attribute on the view that you want to apply this to. That way we maintain "backwards" compatibility with the front end until we're ready to address the other views.

catalogs/models.py
```diff
@@ -1,11 +1,16 @@
...
+from typing import List
 
 import jsonschema
 from django.db import models
+from django.db.models.signals import post_save
 from django.dispatch import receiver
+from django.utils.translation import gettext_lazy as _
 from jsonschema.exceptions import SchemaError, ValidationError
 
+from .catalogio import CatalogTools
```
jwhite3: Non-blocking: We should move away from relative imports and use absolute imports.

catalogs/models.py
```diff
@@ -33,17 +43,50 @@ class Catalog(models.Model):
...
+                control_objects.append(
+                    Controls(
+                        catalog=instance,
+                        control_id=c,
+                        control_label=control_data.get("label"),
+                        sort_id=control_data.get("sort_id"),
+                        title=control_data.get("title"),
+                    )
+                )
+            Controls.objects.bulk_create(control_objects)
```
jwhite3, line 83: Nice

components/tests.py
```diff
@@ -149,7 +149,7 @@ def test_type_max_length(self):
         self.assertEqual(max_length, 100)
 
 
-TEST_COMPONENT_CONTROLS = ["ac-2.1", "ac-6.10", "ac-8", "au-6.1", "sc-2"]
+TEST_COMPONENT_CONTROLS = ["ac-1", "ac-2", "at-1", "at-2", "st-3"]
```
jwhite3, line 152: Why update the test cases here? Does it correspond to the updated test catalog json?

projects/models.py
```diff
@@ -97,18 +106,41 @@ class Meta:
...
+    def save(self, *args, **kwargs):
+        cat = apps.get_model("catalogs", "Catalog")
+        if self.impact_level and self.catalog_version:
+            catalog = cat.objects.get(
+                impact_level=self.impact_level, version=self.catalog_version
+            )
+
+        self.catalog = catalog
+        self.full_clean()
+        super().save(*args, **kwargs)
```
jwhite3, line 118: Now that I think of it, do we want/need this to run every time a project is saved?

projects/tests.py
```diff
@@ -526,26 +563,28 @@ def setUpTestData(cls):
             component_json=TEST_COMPONENT_JSON_BLOB,
         )
 
-    def test_post_save_component_added_ocisco(self):
+    def test_post_save_component_added_ocisco(cls):
```
jwhite3: This isn't a class method, so why change `self` -> `cls`?

catalogs/tests_catalogio.py
```diff
@@ -77,7 +77,7 @@ def test_get_controls_all(self):
     def test_get_controls_all_ids(self):
         controls = self.catalog.get_controls_all_ids()
         self.assertIsInstance(controls, List)
-        self.assertEquals(len(controls), 186)
+        self.assertEquals(len(controls), 53)
```
jwhite3, line 80: Assume it's the test file update that requires this?

projects/views.py
```diff
@@ -155,3 +157,14 @@ def get(self, request, *args, **kwargs):
 
         serializer = self.get_serializer(queryset, many=True)
         return Response(serializer.data, status=status.HTTP_200_OK)
+
+
+class ProjectGetControlList(generics.ListAPIView):
+    queryset = ProjectControl.objects.all().order_by("control_id")
```
jwhite3: So close... but this won't handle the permissions correctly. 

To access a project's controls list, you need to have permission to access the project, but since we have a list view that's meant to provide a particular projects related controls, and the `ListAPIView` was used, permissions on the project aren't checked anywhere (I mentioned/hinted at this being a special case in our brainstorming meeting).

Additionally, the list view only filters the provided QuerySet and doesn't execute `get_object`, so I'm not sure how the right controls are being selected here based on the project id.

jwhite3 (changes requested):

components/tests.py
TMCamp: Yes

projects/models.py
TMCamp: Probably not, but we should make a decision about allowing the user to change their impact level and the ramifications of doing so. If the user changes their impact level we would need to update the Project Controls accordingly, but we wouldn't want to just remove them and replace them if some are marked as completed or in progress.

jwhite3: Hmm, yeah we should come up with a workflow for that scenario.

projects/views.py
```diff
@@ -155,3 +158,26 @@ def get(self, request, *args, **kwargs):
...
+    pagination_class = PageNumberPagination
+
+    def list(self, request, *args, **kwargs):
+        project_id = self.kwargs["project_id"]
+        project = self.get_object()
+        queryset = self.filter_queryset(
+            project.controls.through.objects.filter(project_id=project_id).order_by(
+                "control_id"
+            )
+        )
```
jwhite3: I'm not sure that we need to grab the through table fields in this way, but we can circle back around and improve.

projects/views.py
```diff
@@ -155,3 +158,26 @@ def get(self, request, *args, **kwargs):
...
+        page = self.paginate_queryset(queryset)
```
jwhite3: Also if you want the filters to work you'd need to specify a FilterSet class or filter fields and then add:
```python
    ...
    filtere_qs = self.filter_queryset(queryset)
    page = ...
```
Similar to the comment above, we can improve this later.

jwhite3 (dismissed):

jauschalley (dismissed):
I'm learning a lot from this process.  Thanks for the huge effort here.

jwhite3:
> I'm learning a lot from this process. Thanks for the huge effort here.

That's what I like to hear (read?)!

jwhite3 (approved):

projects/models.py
jwhite3: @TMCamp want to file an issue for this in Jira and we can resolve this conversation?

jauschalley (approved):

rmoore-theta (approved):

projects/models.py
TMCamp: Done: https://jiraent.cms.gov/browse/ISPGBSS-1229

PR #85, opened 2022-08-12
Add linting workflow

jwhite3:
*What does this PR do?*
This PR adds a linter Actions workflow with `pylint`

*Jira ticket number?*
<get ticket number>

*Changes*
- Add `lint.yml`workflow file
- Update `pyproject.toml` with linter dependencies and pylint config

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -17,6 +17,7 @@ jobs:
         run: pip install --upgrade pip
       - name: Install poetry
         run: pip install poetry
+      - run: poetry config virtualenvs.create false
       - name: Set up python 3.9
         uses: actions/setup-python@v4
         with:
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,5 +1,5 @@
 ---
-name: Test
+name: Lint
 on:
   push:
     branches:
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -13,16 +13,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3.0.2
+      - name: Upgrade pip
+        run: pip install --upgrade pip
+      - name: Install poetry
+        run: pip install poetry
       - name: Set up python 3.9
         uses: actions/setup-python@v4
         with:
           python-version: '3.9'
           cache: 'poetry'
-      - name: Upgrade pip
-        run: pip install --upgrade pip
-      - name: Install poetry
-        run: pip install poetry
-      - run: poetry config virtualenvs.create false
       - name: Install dependencies
         run: poetry install --dev --no-ansi --no-interaction --no-root
       - name: Run pylint
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -23,6 +23,6 @@ jobs:
         run: pip install poetry
       - run: poetry config virtualenvs.create false
       - name: Install dependencies
-        run: poetry install --with dev --no-ansi --no-interaction --no-root
+        run: poetry install --no-ansi --no-interaction --no-root
       - name: Run pylint
         run: pylint .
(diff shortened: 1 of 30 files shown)


jwhite3:
To do: 
- [x] Decide on how strict pylint should be
- [x] Update code to satisfy new linting standards

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,26 @@
...
+on:
+  push:
+    branches:
+      - main
+      - develop
+  pull_request:
+
+jobs:
+  lint:
+    name: Pylint
```
jwhite3, line 12: I decided to go with pylint instead of flake8 because it has some useful suggestions/warnings that flake8 doesn't have, but if we want to we can always run both.

jauschalley: Works for me.

jwhite3:
@TMCamp Once this is in we should also configure the listing check to be required.

jauschalley (dismissed):

jwhite3:
@TMCamp ping!

catalogs/catalogio.py
```diff
@@ -5,15 +5,15 @@
...
     """Represent a catalog"""
 
     def __init__(self, source, text=False):
         try:
             self.oscal = self._load_catalog_json(source, text)
-        except (IOError, FileNotFoundError, json.decoder.JSONDecodeError) as e:
-            logger.error(f"Unable to load catalog {source}: {e}")
-            raise CatalogLoadError(f"Could not load catalog {source}") from e
+        except (IOError, FileNotFoundError, json.decoder.JSONDecodeError) as exc:
+            logger.error("Unable to load catalog %s: %s", source, exc)
```
TMCamp, line 15: Is there a reason to not use f strings?

access_management/permission_constants.py
```diff
@@ -28,3 +28,5 @@
     PROJECT_CONTRIBUTOR_GROUP: project_contributor_permissions,
     PROJECT_VIEW_ONLY_GROUP: project_view_only_permissions,
 }
+
+all_perms = vars()
```
TMCamp, line 32: Just for my edification, what is the intent here?

catalogs/catalogio.py
jwhite3: In logging, you usually want to use lazy formatting so that the strings aren't evaluated unless they need to be. Using an f-string literal here would mean that the string is always evaluated.

access_management/permission_constants.py
jwhite3: The intent was to remove the wildcard import over in `access_management.utils`, but I didn't want to significantly refactor this code since it would be out of scope for this PR. Just moved this `dict` here so that we could import elsewhere. 

The use of `vars` in `access_management.utils` was a convenient (although opaque) way of automatically constructing a `dict` that contained these different permssions/groups that could be access programmatically. In a future update we should probably move away from doing this and move a lot of this stuff over to `projects.permissions.py`

jwhite3: `vars` itself just returns the `__dict__` attribute of whatever context it's called in. In this case it's retrieving the `__dict__` attribute of the module.

https://docs.python.org/3/library/functions.html#vars

TMCamp: Awesome. Thank you!

jauschalley: I could've really used this explanation 3 months ago.

jauschalley (approved):
so. many. changes.

jwhite3:
> so. many. changes.

Yeah.. they're mostly "cosmetic" in the sense that they adhere more closely to PEP8 (plus more type hinting and whatnot), but there were some lint "fails" that required more of an update. Luckily this isn't the norm and we shouldn't have to do massive updates like this all the time. Just had to go hand in hand with the introduction of the lint workflow so it would pass.

jauschalley:
> 

I should've used my conventional commenting here and led with 'joke'.  Adding linting is always a marathon if not done right away.  Thanks for adding it.

jwhite3:
> I should've used my conventional commenting here and led with 'joke'. Adding linting is always a marathon if not done right away. Thanks for adding it.

No worries, I new it wasn't a "serious" comment, but I still feel bad for it lol

TMCamp (approved):

PR #86, opened 2022-08-15
Bug/fix component types filter

jwhite3:
*What does this PR do?*
Fixes a bug in the component types view where the incorrect permissions filter was being used resulting in missing types in the response.

*Jira ticket number?*
[ISPGBSS-1219](https://jiraent.cms.gov/browse/ISPGBSS-1219)

*Changes*
- Update `ComponentTypeListView` to use the more specific `ComponentPermissionsFilter` class.
- Update `ComponentPermissionsFilter` QuerySet to use a `union` instead of the `|` operator

### DIFF
--- a/components/filters.py
+++ b/components/filters.py
@@ -37,5 +37,5 @@ class ComponentPermissionsFilter(BaseFilterBackend):
 
         return (
                 get_objects_for_user(user, permission, queryset, accept_global_perms=False)
-                | queryset.filter(status=Component.Status.PUBLIC)
+                .union(queryset.filter(status=Component.Status.PUBLIC))
         )
--- a/components/filters.py
+++ b/components/filters.py
@@ -38,4 +38,5 @@ class ComponentPermissionsFilter(BaseFilterBackend):
         return (
                 get_objects_for_user(user, permission, queryset, accept_global_perms=False)
                 .union(queryset.filter(status=Component.Status.PUBLIC))
+                .order_by("pk")
         )
--- a/components/tests.py
+++ b/components/tests.py
@@ -502,8 +502,11 @@ class ComponentTypesViewTest(AuthenticatedAPITestCase):
     def test_get_types_list(self):
         resp = self.client.get("/api/components/types/", format="json")
         self.assertEqual(resp.status_code, 200)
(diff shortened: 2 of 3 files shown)


rmoore-theta (dismissed):

jauschalley (dismissed):

components/filters.py
```diff
@@ -37,5 +37,6 @@ def filter_queryset(self, request, queryset, view):
 
         return (
                 get_objects_for_user(user, permission, queryset, accept_global_perms=False)
-                | queryset.filter(status=Component.Status.PUBLIC)
+                .union(queryset.filter(status=Component.Status.PUBLIC))
+                .order_by("pk")
```
jwhite3, line 41: It's necessary to provide the `order_by` here since `union` doesn't preserve any `order_by` in the individual query sets.

components/tests.py
```diff
@@ -500,8 +502,38 @@ def setUpTestData(cls):
     def test_get_types_list(self):
         resp = self.client.get("/api/components/types/", format="json")
         self.assertEqual(resp.status_code, 200)
-        self.assertEqual(json.loads(resp.content)[0][0], "policy")
-        self.assertEqual(json.loads(resp.content)[1][0], "software")
+
+        content = resp.json()
+        # Data structure in response is [["type1"], ["type2"], ..]
+        flattened = sorted(value for item in content for value in item)
+        self.assertEqual(flattened, ["policy", "software"])
```
jwhite3, line 509: Updated the test to be more values based rather than order based to prevent this test from failing due to order changes.

components/views.py
```diff
@@ -14,7 +13,7 @@ class ComponentListView(generics.ListCreateAPIView):
     """Use for read-write endpoints to represent a collection of model instances.
     Provides get and post method handlers.
     """
-    queryset = Component.objects.all().order_by("pk")
+    queryset = Component.objects.all()
```
jwhite3, line 16: Ordering is occurring in the permissions filter.

components/tests.py
rmoore-theta: :) I like this approach much better

rmoore-theta (approved):

TMCamp (approved):

PR #88, opened 2022-08-16
Bug/Fix duplicate types in component types endpoint

jwhite3:
*What does this PR do?*
This PR fixes a bug in the component types view that results in duplicated types.

*Jira ticket number?*
[ISPGBSS-1220](https://jiraent.cms.gov/browse/ISPGBSS-1220)

*Changes*
- Update test to include a component with a duplicate type

### DIFF
--- a/components/filters.py
+++ b/components/filters.py
@@ -38,5 +38,4 @@ class ComponentPermissionsFilter(BaseFilterBackend):
         return (
                 get_objects_for_user(user, permission, queryset, accept_global_perms=False)
                 .union(queryset.filter(status=Component.Status.PUBLIC))
-                .order_by("pk")
         )
--- a/components/filters.py
+++ b/components/filters.py
@@ -36,6 +36,6 @@ class ComponentPermissionsFilter(BaseFilterBackend):
         }
 
         return (
-                get_objects_for_user(user, permission, queryset, accept_global_perms=False)
-                .union(queryset.filter(status=Component.Status.PUBLIC))
+            get_objects_for_user(user, permission, queryset, accept_global_perms=False)
+            .union(queryset.filter(status=Component.Status.PUBLIC))
         )
--- a/components/views.py
+++ b/components/views.py
@@ -1,4 +1,5 @@
 from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
+from django.db.models import QuerySet
 from django_filters import rest_framework as filters
 from rest_framework import generics, status
 from rest_framework.response import Response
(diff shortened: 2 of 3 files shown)


components/filters.py
```diff
@@ -38,5 +38,4 @@ def filter_queryset(self, request, queryset, view):
         return (
                 get_objects_for_user(user, permission, queryset, accept_global_perms=False)
                 .union(queryset.filter(status=Component.Status.PUBLIC))
-                .order_by("pk")
```
jwhite3, line 41: I thought about this some more, and I didn't think this was the best place for the `order_by`.

components/views.py
```diff
@@ -56,11 +57,11 @@ def list(self, request, *args, **kwargs):
...
+    queryset = Component.objects.values_list("type")
     permission_classes = [ComponentPermissions, ]
     filter_backends = [ComponentPermissionsFilter, ]
 
-    def list(self, request, *args, **kwargs):
+    def list(self, request: Request, *args, **kwargs) -> Response:
         queryset = self.filter_queryset(self.get_queryset())
 
-        return Response(queryset, status=status.HTTP_200_OK)
+        return Response(set(queryset), status=status.HTTP_200_OK)
```
jwhite3: `distinct` can't be applied to a `union` QuerySet.

TMCamp (dismissed):

rmoore-theta (dismissed):

components/views.py
```diff
@@ -56,11 +62,11 @@ def list(self, request, *args, **kwargs):
 
 
 class ComponentTypeListView(generics.ListAPIView):
-    queryset = Component.objects.order_by().values_list("type").distinct()
+    queryset = Component.objects.values_list("type")
```
jwhite3, line 65: Not entirely sure what was happening here to cause duplicates after applying the `union`, but removing the `order_by` and `distinct` calls fixes the issue. They weren't really useful in this case anyways.

components/views.py
```diff
@@ -18,6 +20,10 @@ class ComponentListView(generics.ListCreateAPIView):
     permission_classes = [ComponentPermissions, ]
     filter_backends = [ComponentPermissionsFilter, ]
 
+    def filter_queryset(self, queryset: QuerySet) -> QuerySet:
+        # Need to apply the order_by after the union in ComponentPermissionsFilter
+        return super().filter_queryset(queryset).order_by("pk")
+
```
jwhite3, line 26: Even though the tests passed, they could randomly fail because the `order_by` wasn't being preserved. Needs to be applied after the `union`, but I didn't think the `ComponentPermissionsFilter` applying an ordering that couldn't be changed was a good idea. Might need to think about this some more in the future, but I think this is fine for now.

rmoore-theta (approved):

jauschalley (approved):

PR #89, opened 2022-08-19
Add endpoint for updating a component's control narrratives

rmoore-theta:
*What does this PR do?*
Allows for updating components by adding/editing a control narrative
*Jira ticket number?*
[ISPGBSS-1172](https://jiraent.cms.gov/browse/ISPGBSS-1172)

*Changes*
- Added new serializer for use in updating a components control narrative
- Added new update view

### DIFF
--- a/components/permissions.py
+++ b/components/permissions.py
@@ -7,8 +7,20 @@ from components.models import Component
 
 
 class ComponentPermissions(StrictDjangoObjectPermissions):
-    def has_object_permission(self, request: Request, view: APIView, obj: Component) -> bool:
+    def has_object_permission(
+        self, request: Request, view: APIView, obj: Component
+    ) -> bool:
         if obj.status == Component.Status.PUBLIC and request.method in SAFE_METHODS:
+            return True
+
+        return super().has_object_permission(request, view, obj)
+
+
+class ComponentPrivatePermissions(StrictDjangoObjectPermissions):
+    def has_object_permission(
+        self, request: Request, view: APIView, obj: Component
+    ) -> bool:
+        if request.method in SAFE_METHODS:
             return True
 
         return super().has_object_permission(request, view, obj)
--- a/components/permissions.py
+++ b/components/permissions.py
@@ -14,13 +14,3 @@ class ComponentPermissions(StrictDjangoObjectPermissions):
             return True
 
         return super().has_object_permission(request, view, obj)
-
-
-class ComponentPrivatePermissions(StrictDjangoObjectPermissions):
-    def has_object_permission(
-        self, request: Request, view: APIView, obj: Component
-    ) -> bool:
-        if request.method in SAFE_METHODS:
-            return True
-
-        return super().has_object_permission(request, view, obj)
--- a/components/serializers.py
+++ b/components/serializers.py
@@ -78,30 +78,56 @@ class ComponentSerializer(serializers.ModelSerializer):
         )
 
     def partial_update(self, instance, validated_data):
-        if controls := validated_data.get("controls"):  # && controls not in:
-            print(controls)
-            # instance.controls.append(controls)
-            instance.controls = list(set(instance.controls).union(controls))
+        if controls := validated_data.get("controls"):
+            existing_control = controls not in instance.controls
+            if not existing_control:
+                instance.controls = list(set(instance.controls).union(controls))
 
-        if description := validated_data.get("description"):
-            implemented_requirement = {
-                "uuid": str(uuid.uuid4()),
-                "props": [
-                    {
-                        "name": "security_control_type",
-                        "uuid": str(uuid.uuid4()),
-                        "value": "Hybrid",
-                    },
-                    {"name": "provider", "uuid": str(uuid.uuid4()), "value": "No"},
-                ],
-                "control-id": controls[0],
-                "description": description,
-            }
-            instance.component_json.get("component-definition").get("components")[
-                0
-            ].get("control-implementations")[0].get("implemented-requirements").append(
-                implemented_requirement
-            )
+            if description := validated_data.get("description"):
+                implemented_requirement = {
+                    "uuid": str(uuid.uuid4()),
+                    "props": [
+                        {
+                            "name": "security_control_type",
+                            "uuid": str(uuid.uuid4()),
+                            "value": "Hybrid",
+                        },
+                        {"name": "provider", "uuid": str(uuid.uuid4()), "value": "No"},
+                    ],
+                    "control-id": controls[0],
+                    "description": description,
+                }
+                if existing_control:
+                    for implemented in (
+                        instance.component_json.get("component-definition")
+                        .get("components")[0]
+                        .get("control-implementations")[0]
+                        .get("implemented-requirements")
+                    ):
+                        if implemented.get("control-id") == controls[0]:
+                            instance.component_json.get("component-definition").get(
+                                "components"
+                            )[0].get("control-implementations")[0].get(
+                                "implemented-requirements"
+                            ).remove(
+                                implemented
+                            )
+
+                    instance.component_json.get("component-definition").get(
+                        "components"
+                    )[0].get("control-implementations")[0].get(
+                        "implemented-requirements"
+                    ).append(
+                        implemented_requirement
+                    )
+                else:
+                    instance.component_json.get("component-definition").get(
+                        "components"
+                    )[0].get("control-implementations")[0].get(
+                        "implemented-requirements"
+                    ).append(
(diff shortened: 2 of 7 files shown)


components/permissions.py
```diff
@@ -7,8 +7,20 @@
...
+    def has_object_permission(
+        self, request: Request, view: APIView, obj: Component
+    ) -> bool:
         if obj.status == Component.Status.PUBLIC and request.method in SAFE_METHODS:
             return True
 
         return super().has_object_permission(request, view, obj)
+
+
+class ComponentPrivatePermissions(StrictDjangoObjectPermissions):
```
rmoore-theta: Not sure if this is really needed, but I added it to make sure private components are not filtered out by 
if obj.status == Component.Status.PUBLIC

components/serializers.py
```diff
@@ -72,8 +74,38 @@ class Meta:
             "catalog_data",
             "component_data",
             "project_data",
+            "component_json",
         )
 
+    def partial_update(self, instance, validated_data):
```
rmoore-theta: Ignore for now, going to try to utilize this as a patch call instead of a new endpoint with a put...

components/serializers.py
```diff
@@ -143,6 +175,10 @@ def collect_project_data(component_id, user):
     return form_values
 
 
+def add_control_narrative():
```
rmoore-theta: pretty sure this will not be used...

components/serializers.py
```diff
@@ -159,3 +195,92 @@ class Meta:
...
+            print("controls", controls)
+            print("existing controls", instance.controls)
+            existing_control = controls not in instance.controls
+            print("existing_control", existing_control)
+            # instance.controls.append(controls)
+            if not existing_control:
+                instance.controls = list(set(instance.controls).union(controls))
+
+            print(controls)
+            if description := validated_data.get("description"):
```
rmoore-theta: I nested the description inside of the controls because we need the controls to get the id we are changing the description for... might be able to combined them :/

components/serializers.py
```diff
@@ -159,3 +195,92 @@ class Meta:
...
+                }
+                if existing_control:
+                    for implemented in (
+                        instance.component_json.get("component-definition")
+                        .get("components")[0]
+                        .get("control-implementations")[0]
+                        .get("implemented-requirements")
+                    ):
+                        # form_values["add"].append({"value": a.id, "label": a.title})
+                        if implemented.get("control-id") == controls[0]:
```
rmoore-theta: @jwhite3 For some reason this only removes 1 "implemented-requirements" with a matching control-id, which we should only have 1, but when I broke the data and have a duplicate, it still leaves a duplicate :/

rmoore-theta: Python ends the loop on successful if condition... This is fine as long as we keep data clean.

components/serializers.py
```diff
@@ -72,8 +74,38 @@ class Meta:
...
 
+    def partial_update(self, instance, validated_data):
+        if controls := validated_data.get("controls"):  # && controls not in:
+            print(controls)
+            # instance.controls.append(controls)
+            instance.controls = list(set(instance.controls).union(controls))
+
+        if description := validated_data.get("description"):
+            implemented_requirement = {
+                "uuid": str(uuid.uuid4()),
```
TMCamp: You should be using the Pydantic models in `/blueprintapi/oscal/`

TMCamp: You can reference the EmptyComponent class in `componentio.py`

components/permissions.py
```diff
@@ -7,8 +7,20 @@
 
 
 class ComponentPermissions(StrictDjangoObjectPermissions):
-    def has_object_permission(self, request: Request, view: APIView, obj: Component) -> bool:
+    def has_object_permission(
+        self, request: Request, view: APIView, obj: Component
+    ) -> bool:
```
jwhite3, line 12: I'm starting to hate `black` lol

components/permissions.py
jwhite3: No, I don't think this is needed.

The other one effectively doesn't check for any particular perms for public components on "safe" (GET, OPTIONS, HEAD), but for private ones it does. If used with the permissions filter, this means that only private components that the current user doesn't have access to will be filtered out.

components/serializers.py
```diff
@@ -72,8 +74,38 @@ class Meta:
...
+                "props": [
+                    {
+                        "name": "security_control_type",
+                        "uuid": str(uuid.uuid4()),
+                        "value": "Hybrid",
+                    },
+                    {"name": "provider", "uuid": str(uuid.uuid4()), "value": "No"},
```
jwhite3: Each time you call `uuid.uuid4` you'll get a different id. Was it intentional that each of these fields had a separate uuid?

components/serializers.py
jwhite3: I think this might be easier once #84 is in, but I think it would be better to raise an error here if `control_id` is required, but isn't present. 

Does beg the question though: what is `instance` in this case? A project?

components/serializers.py
```diff
@@ -159,3 +195,70 @@ class Meta:
...
+            if description := validated_data.get("description"):
+                implemented_requirement = {
+                    "uuid": str(uuid.uuid4()),
+                    "props": [
+                        {
+                            "name": "security_control_type",
+                            "uuid": str(uuid.uuid4()),
+                            "value": "Hybrid",
+                        },
+                        {"name": "provider", "uuid": str(uuid.uuid4()), "value": "No"},
```
jwhite3: Same question as above: is it intentional for each of these "uuid" fields to be different ids?

components/views.py
```diff
@@ -5,35 +5,54 @@
...
+
     queryset = Component.objects.all()
-    permission_classes = [ComponentPermissions, ]
+    permission_classes = [
+        ComponentPermissions,
+    ]
     serializer_class = ComponentSerializer
 
+    def patch(self, request, *args, **kwargs):
+        return self.partial_update(request, *args, **kwargs)
```
jwhite3: This looks suspiciously like the implementation already in `RetrieveUpdateAPIView`. Probably can scrap it unless you want to extend or override it?

components/views.py
```diff
@@ -57,10 +76,28 @@ def list(self, request, *args, **kwargs):
...
+    permission_classes = [
+        ComponentPrivatePermissions,
+    ]
+    serializer_class = ComponentControlSerializer
+
+    # def patch(self, request, *args, **kwargs):
+    #     return self.partial_update(request, *args, **kwargs)
+
+    def patch(self, request, *args, **kwargs):
+        return self.update(request, *args, **kwargs)
```
jwhite3: I don't think it's a good idea mixing and matching like this. Everywhere else where we use `generics`, a patch will allow partial updates, but this won't.

jwhite3 (commented):
We should meet up to chat about the update method in `ComponentControlSerializer`. I think I need more context there.

components/permissions.py
rmoore-theta: lol

components/serializers.py
rmoore-theta: `instance` is a component

components/serializers.py
rmoore-theta: Yes, these uuids are intentionally different. This is the reference material it is based on https://github.com/CMSgov/component-library/blob/main/django/oscal/django.json

components/serializers.py
```diff
@@ -159,3 +160,61 @@ class Meta:
...
+        Property(name="provider", value="no"),
+    ]
+    ir = ImplementedRequirement(
+        props=props, control_id=controls[0], description=description
+    )
+    print("ir", ir)
+    instance.component_json.get("component-definition").get("components")[0].get(
+        "control-implementations"
+    )[0].get("implemented-requirements").append(ir)
+
```
Tom-Camp: I think that when you append the ImplementedRequirements you need to convert it JSON since you are trying to append a Pydantic object to a JSON string. Something like `.append(json.loads(IR.json()))`

components/serializers.py
```diff
@@ -159,3 +162,68 @@ class Meta:
             "catalog",
             "controls_count",
         )
+
+
+class ComponentControlSerializer(serializers.ModelSerializer):
+    # catalog_name = serializers.SerializerMethodField()
```
jwhite3: This should probably be a `CharField` instead of a `SerializerMethodField`. With a `CharField` you shouldn't need the `get_catalog_name` anymore.

TMCamp (approved):

jauschalley (approved):

PR #91, opened 2022-08-22
Feature/update individual controls

jwhite3:
*What does this PR do?*
This PR provides support for updating the `status` field for controls associated with a project.

*Jira ticket number?*
[ISPGBSS-1195](https://jiraent.cms.gov/browse/ISPGBSS-1195)

*Changes*
- Update project control serializer (consolidate/update existing)
- Add new project control view
- Add new tests
- Clean up project tests
- Add `load_catalog` management command

### DIFF
--- a/catalogs/management/commands/load_catalog.py
+++ b/catalogs/management/commands/load_catalog.py
@@ -27,8 +27,8 @@ class Command(BaseCommand):
                 create_kwargs = {"file_name": file, "name": name}
 
                 for arg in ("source", "catalog_version", "impact_level"):
-                    if arg := options[arg]:
-                        create_kwargs[arg if arg != "catalog_version" else "version"] = arg
+                    if value := options[arg]:
+                        create_kwargs[arg if arg != "catalog_version" else "version"] = value
 
                 Catalog.objects.create(**create_kwargs)
 
--- a/catalogs/management/commands/load_catalog.py
+++ b/catalogs/management/commands/load_catalog.py
@@ -1,0 +1,40 @@
+from pathlib import Path
+
+from django.core.files import File
+from django.core.management.base import BaseCommand, CommandError
+from django.db import IntegrityError
+
+from catalogs.models import Catalog
+
+
+class Command(BaseCommand):
+    help = "Ingest catalog data into the Catalog and Control tables."
+
+    def add_arguments(self, parser):
+        parser.add_argument("--catalog-file", type=str)
+        parser.add_argument("--impact-level", type=str, default=None)
+        parser.add_argument("--name", type=str)
+        parser.add_argument("--source", type=str, default=None)
+        parser.add_argument("--catalog-version", type=str, default=None)
+
+    def handle(self, *args, **options):
+        input_file = Path(options["catalog_file"])
+
+        try:
+            with open(input_file) as catalog:
+                file = File(catalog)
+                name = input_name if (input_name := options["name"]) else input_file.name
+                create_kwargs = {"file_name": file, "name": name}
+
+                for arg in ("source", "catalog_version", "impact_level"):
+                    if arg := options[arg]:
+                        create_kwargs[arg if arg != "catalog_version" else "version"] = arg
+
+                Catalog.objects.create(**create_kwargs)
+
+        except IntegrityError as exc:
+            raise CommandError(f"Error in creating new catalog: {exc}") from exc
+        except (IOError, FileNotFoundError) as exc:
+            raise CommandError(f"Error loading catalog file: {exc}") from exc
+
+        self.stdout.write(self.style.SUCCESS('Successfully ingested catalog "%s"' % name))
--- a/catalogs/management/commands/load_catalog.py
+++ b/catalogs/management/commands/load_catalog.py
@@ -21,7 +21,7 @@ class Command(BaseCommand):
         input_file = Path(options["catalog_file"])
 
         try:
-            with open(input_file) as catalog:
+            with open(input_file, mode="rb") as catalog:
                 file = File(catalog)
                 name = input_name if (input_name := options["name"]) else input_file.name
                 create_kwargs = {"file_name": file, "name": name}
--- a/projects/permissions.py
+++ b/projects/permissions.py
@@ -1,0 +1,36 @@
+from django.http import Http404
+from rest_framework.permissions import SAFE_METHODS
+
+from blueprintapi.permissions import StrictDjangoObjectPermissions
+
+
+class ProjectControlPermissions(StrictDjangoObjectPermissions):
+    """Override StrictDjangoObjectPermissions to check required permissions based on the obj class instead of the view's
+    queryset.
+    This is specific to project control where ProjectControl object permissions should be effectively "inherited" based
+    on a user's project permission.
+    """
+    def has_object_permission(self, request, view, obj):
+        model_cls = obj.__class__
+        user = request.user
+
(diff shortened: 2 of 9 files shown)


catalogs/management/commands/load_catalog.py
```diff
@@ -0,0 +1,40 @@
+from pathlib import Path
+
+from django.core.files import File
+from django.core.management.base import BaseCommand, CommandError
+from django.db import IntegrityError
+
+from catalogs.models import Catalog
+
+
+class Command(BaseCommand):
```
jwhite3, line 10: This adds a "management" command that can be run with: 
```
python manage.py load_catalog
```
options: 
```
usage: manage.py load_catalog [-h] [--catalog-file CATALOG_FILE] [--impact-level IMPACT_LEVEL] [--name NAME] [--source SOURCE] [--catalog-version CATALOG_VERSION] [--version] [-v {0,1,2,3}] [--settings SETTINGS]
                              [--pythonpath PYTHONPATH] [--traceback] [--no-color] [--force-color] [--skip-checks]

Ingest catalog data into the Catalog and Control tables.

optional arguments:
  -h, --help            show this help message and exit
  --catalog-file CATALOG_FILE
  --impact-level IMPACT_LEVEL
  --name NAME
  --source SOURCE
  --catalog-version CATALOG_VERSION
  --version             Show program's version number and exit.
  -v {0,1,2,3}, --verbosity {0,1,2,3}
                        Verbosity level; 0=minimal output, 1=normal output, 2=verbose output, 3=very verbose output
  --settings SETTINGS   The Python path to a settings module, e.g. "myproject.settings.main". If this isn't provided, the DJANGO_SETTINGS_MODULE environment variable will be used.
  --pythonpath PYTHONPATH
                        A directory to add to the Python path, e.g. "/home/djangoprojects/myproject".
  --traceback           Raise on CommandError exceptions.
  --no-color            Don't colorize the command output.
  --force-color         Force colorization of the command output.
  --skip-checks         Skip system checks.
```

projects/serializers.py
```diff
@@ -20,6 +22,7 @@ class Meta:
             "location",
             "catalog",
         )
+        read_only_fields = ("id", "creator")
         extra_kwargs = {"creator": {"read_only": True}}
```
jwhite3: Forgot to remove this.

projects/tests.py
```diff
@@ -97,132 +99,32 @@ def setUpTestData(cls):
             catalog_version="NIST 800-53",
             impact_level="low",
             location="other",
-            creator=cls.test_user,
-            catalog=cls.test_catalog,
-        )
-
-    # Tests for field labels
```
jwhite3, line 104: These were removed for being too fine-grained. At this point it's just testing django, which is unnecessary. That being said, if folks spot one that shouldn't have been removed that I removed by accident, let me know.

projects/tests.py
```diff
@@ -750,16 +661,103 @@ def test_ociso_component_not_returned(self):
...
+            location="other",
+            creator=user,
+            catalog=test_catalog,
+        )
+
+    def setUp(self):
+        self.client.force_authenticate(user=self.user, token=self.token)
+
+    def test_get_project_control(self):
+        response = self.client.get(
```
jwhite3, line 721: Test currently fails because the backend checks for permissions based on the queryset model (`ProjectControl`) rather than what we want (`Project`). I'm working on figuring out the best way to tackle this since it's fairly common in this project to access a list of things associated with a particular project.

projects/serializers.py
```diff
@@ -20,7 +22,7 @@ class Meta:
             "location",
             "catalog",
         )
-        extra_kwargs = {"creator": {"read_only": True}}
+        read_only_fields = ("id", "creator", "catalog", )
```
jwhite3, line 25: Let me know if this is something that's not possible. At the moment, `Project.save` reassigns catalog based on impact level and catalog version, so passing a catalog id wasn't actually doing anything anyways.

projects/tests.py
```diff
@@ -97,240 +99,150 @@ def setUpTestData(cls):
...
-                has_default = True
 
-        self.assertTrue(has_default)
+    def test_project_has_default_component(self):
+        private_component = self.test_project.components.get(title="Pretty Ordinary Project private")
+        self.assertEqual(private_component.status, Component.Status.SYSTEM)
 
 
-class ProjectRequiredFieldsTest(AuthenticatedAPITestCase):
+class ProjectCreateViewTestCase(AuthenticatedAPITestCase):
```
jwhite3, line 116: I noticed that we didn't have a test that actually tried to create a project with valid data, so I refactored/extended this test case to include a test for attempting to create a "legit" project while consolidating the "required fields" tests into a single, parametrized test.

catalogs/management/commands/load_catalog.py
jauschalley: Thanks for this.

projects/serializers.py
jauschalley: This makes sense to me but I'd rely on Tom's input.

jwhite3: @TMCamp thoughts?

projects/serializers.py
```diff
@@ -52,61 +54,55 @@ class Meta:
...
         component_data = {
             "responsibility": "Allocated",
             "components": {"inherited": {}, "private": {}},
         }
         count = 0
         responsibility: str = ""
-        for c in components.all():
-            status = "inherited" if c.status == 2 else "private"
-            if control_id in c.controls:
+        for component in components:
```
jauschalley, line 98: Praise: I'm a huuuuuge fan of not using single letter variables.  I don't typically block PRs over it but I definitely approve the improvements 😉

projects/serializers.py
```diff
@@ -52,61 +54,55 @@ class Meta:
...
-            "title",
-            "acronym",
-            "catalog_data",
-            "component_data",
-        )
+        return {}
 
-    def get_control_data(self, components, control_id):
+    @staticmethod
+    def _get_control_data(components: QuerySet, control_id: str) -> dict:
```
jauschalley, line 91: Praise: After looking up why you would use a leading underscore here, I am sure that we have some cleanup to do.  Thanks for setting us straight.

jauschalley (approved):
Looks good to me.

projects/permissions.py
```diff
@@ -0,0 +1,36 @@
...
+            # a 404 response.
+
+            if request.method in SAFE_METHODS:
+                # Read permissions already checked and failed, no need
+                # to make another lookup.
+                raise Http404
+
+            read_perms = self.get_required_object_permissions('GET', model_cls)
+            if not user.has_perms(read_perms, obj):
+                raise Http404
```
rmoore-theta, line 31: Should this one be Http403?

jwhite3: Good question! No, to be consistent with everything else it should be a 404. 

The idea is that if a user doesn't have object-level permissions to even view the object, they shouldn't even know that it exists in the first place. This is how rest framework implemented permissions checks, and is also a common practice more broadly (also more secure). 

Other real world example: if you try and access a private repository on GitHub that you don't have access to you get a 404 "Not Found"

rmoore-theta (approved):

projects/serializers.py
TMCamp: Yep, that makes sense, @jwhite3.

TMCamp (approved):

PR #93, opened 2022-08-23
Feature/add percent complete

jwhite3:
*What does this PR do?*
This PR adds two new fields, `completed_controls` and `total_controls` to the project list and detail view.

*Jira ticket number?*
[ISPGBSS-1079](https://jiraent.cms.gov/browse/ISPGBSS-1079)

*Changes*
- Added annotated QuerySet to `ProjectsListViews` and `ProjectsDetailView`
- Added `completed_controls` and `total_controls` fields to `ProjectListSerializer` and `ProjectSerializer`
- Added tests

### DIFF
--- a/projects/serializers.py
+++ b/projects/serializers.py
@@ -1,5 +1,3 @@
-import decimal
-
 from django.db.models import QuerySet
 from rest_framework import serializers
 
@@ -12,9 +10,8 @@ from projects.models import Project, ProjectControl
 
 
 class ProjectListSerializer(serializers.ModelSerializer):
-    percent_complete = serializers.DecimalField(
-        max_digits=3, decimal_places=0, coerce_to_string=False, required=False, rounding=decimal.ROUND_UP
-    )
+    completed_controls = serializers.IntegerField()
+    total_controls = serializers.IntegerField()
 
     class Meta:
         model = Project
@@ -27,9 +24,10 @@ class ProjectListSerializer(serializers.ModelSerializer):
             "creator",
             "location",
             "catalog",
-            "percent_complete",
+            "completed_controls",
+            "total_controls",
         )
-        read_only_fields = ("id", "creator", "catalog", "percent_complete", )
+        read_only_fields = ("id", "creator", "catalog", "completed_controls", "total_controls", )
 
     def create(self, validated_data):
         return Project.objects.create(
@@ -40,9 +38,8 @@ class ProjectListSerializer(serializers.ModelSerializer):
 class ProjectSerializer(serializers.ModelSerializer):
     components = ComponentListSerializer(many=True)
     components_count = serializers.SerializerMethodField()
-    percent_complete = serializers.DecimalField(
-        max_digits=3, decimal_places=0, coerce_to_string=False, required=False, rounding=decimal.ROUND_UP
-    )
+    completed_controls = serializers.IntegerField()
+    total_controls = serializers.IntegerField()
 
     def get_components_count(self, obj):
         return obj.components.count()
@@ -60,9 +57,10 @@ class ProjectSerializer(serializers.ModelSerializer):
             "components",
             "components_count",
             "catalog",
-            "percent_complete",
+            "completed_controls",
+            "total_controls",
         )
-        read_only_fields = ("id", "creator", "percent_complete", )
+        read_only_fields = ("id", "creator", "completed_controls", "total_controls", )
         depth = 1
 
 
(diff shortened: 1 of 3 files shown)


projects/views.py
```diff
@@ -19,17 +21,27 @@
...
 
+n_completed = Cast(Count("to_project", filter=Q(to_project__status="complete")), FloatField())
+total_controls = Cast(Count("to_project"), FloatField())
+
+project_queryset = (
+        Project
+        .objects
+        .annotate(percent_complete=(n_completed / total_controls) * 100)
+        .order_by("pk")
+    )
```
jwhite3: Annotating the QuerySet I think should be more efficient that implementing this calculation as a property, which may be important when there are a larger number of projects

rmoore-theta (dismissed):

projects/serializers.py
```diff
@@ -33,6 +36,7 @@ def create(self, validated_data):
 class ProjectSerializer(serializers.ModelSerializer):
     components = ComponentListSerializer(many=True)
     components_count = serializers.SerializerMethodField()
+    percent_complete = serializers.DecimalField(max_digits=3, decimal_places=1, coerce_to_string=False, required=False)
```
jauschalley: Non-blocking: I'm not totally sure if we want to do this formatting on the front end or the back end but I think we are going to display the two digit percent format in the UI (ie 67%).  I'd love others' input on this.

TMCamp (dismissed):

projects/serializers.py
jwhite3: I interpreted the ticket as saying that the percentage was meant to be a new field in the API response, but I definitely could've misinterpreted.

For something like this I'm not actually sure if it matters too much, but pros and cons for having it on the backend:
pros: 
- Ship less data to the front end
- Calculation is "refreshed" for each request since it's an annotation on the query set, and rest framework ensures that the query set is reevaluated for each request. Means the result should never be stale as someone works and goes back and forth between individual control pages and the project page. May be less complex than implementing something similar on the front end which may require additional requests.

cons:
- Evaluating the project query set may be slightly slower


In terms of the actual percentage, I guessed for the type and number of decimal places that I thought might be useful, but if we'd prefer not to have decimals that's easy enough!

jauschalley: I agree with all of the above.  I think dropping the decimal would just clean it up a bit since we won't be using it.  We don't need an exact number since it's just a rough indicator and won't be tied to any sort of calculation in the UI.  But still, I'm not attempting to block this so feel free to merge if you're good with it.

jwhite3: @TMCamp @rmoore-theta @amyd11 Thoughts on this?

rmoore-theta: I would default to HCD for if they want it to display. I personally think decimals will be nice if you have a lot of controls you need to complete, but probably overcomplicates things.

jwhite3: @jauschalley in https://github.com/CMSgov/blueprint_api/pull/93/commits/c395bc9347bf9f898b99c15647232ab96a5d2eab I've ensured that the value is rounded up correctly (just casting as an integer truncated the decimal rather than rounding correctly). On the front end, it could be cast as an int or I could send it as a string and it could be stripped?

jwhite3: (also this is super easy to adjust, so we can always see how we like it to start and tweak it later)

jauschalley (dismissed):

projects/serializers.py
amyd11: Hmm, I do think it's nice that the backend already has it calculated, but for the front end we're going to need a few numbers, not just the percent. _(see [this figma design](https://www.figma.com/file/9a26GjOwssW76cTr01hfXL/Guidelines-%26-Wireframes?node-id=1635%3A3984))_.

The front end will need to display:

- total number of controls
- number of controls completed
- percent of controls completed

So, the backend will at least need to send over the total number and the number completed. If the backend also calculates the percent, that could take the calculation off the front end, but the front end would be capable of doing it since it'll have the total and number of complete controls.

amyd11: It looks like we overlooked that detail in the JIRA ticket, unfortunately

jwhite3: Well in that case, ignore everything I said 😭

jwhite3: If you need the completed number and the total might as well calculate the percentage on the front end

jauschalley (dismissed):

jwhite3:
@TMCamp @rmoore-theta @jauschalley @amyd11 Ping!

rmoore-theta (approved):

jauschalley (approved):
The PR that keeps on giving!

PR #95, opened 2022-08-26
589: Updating logging formats. 

TMCamp:
*What does this PR do?*

*Jira ticket number?*
[ISPGBSS-589](https://jiraent.cms.gov/browse/ISPGBSS-589)

*Changes*

- Updated the logging output to include the time that the log entry was created, the logging level, and the message
- Added logging for failed login attempts.

*Testing*

- [ ] Start your local application with the output being shown in the terminal
- [ ] Try to log in with the wrong credentials
- [ ] You should see a message similar to `[2022-08-26 17:37:44,327] [WARNING] Log in failed for: USERNAME`

### DIFF
--- a/blueprintapi/settings.py
+++ b/blueprintapi/settings.py
@@ -34,7 +34,7 @@ LOGGING = {
         },
     },
     "loggers": {
-        "root": {
+        "": {
             "level": "INFO",
             "handlers": ["console"],
         }
--- a/blueprintapi/settings.py
+++ b/blueprintapi/settings.py
@@ -34,7 +34,7 @@ LOGGING = {
         },
     },
     "loggers": {
-        "": {
+        "root": {
             "level": "INFO",
             "handlers": ["console"],
         }
--- a/blueprintapi/settings.py
+++ b/blueprintapi/settings.py
@@ -12,6 +12,7 @@ https://docs.djangoproject.com/en/3.2/ref/settings/
 
 import logging
 import os
+import sys
 from pathlib import Path
 
 from .Environment import Environment
@@ -22,11 +23,21 @@ LOGGING = {
     "handlers": {
         "console": {
             "class": "logging.StreamHandler",
+            "stream": sys.stdout,
+            "formatter": "simple",
         },
     },
-    "root": {
-        "handlers": ["console"],
-        "level": "WARNING",
+    "formatters": {
+        "simple": {
+            "format": "[{asctime}] [{levelname}] {message}",
+            "style": "{",
+        },
+    },
+    "loggers": {
+        "": {
+            "level": "INFO",
+            "handlers": ["console"],
+        }
     },
 }
 
--- a/users/apps.py
+++ b/users/apps.py
@@ -1,11 +1,20 @@
 from django.apps import AppConfig
+from django.contrib.auth.signals import user_login_failed
 from django.db.models.signals import post_save
+
 
 class UsersConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     name = "users"
 
     def ready(self):
-        from users.signals import add_default_user_perms
+        from users.signals import add_default_user_perms, user_login_failed_callback
 
-        post_save.connect(add_default_user_perms, sender="users.User", dispatch_uid="add_user_perms")
+        post_save.connect(
+            add_default_user_perms, sender="users.User", dispatch_uid="add_user_perms"
+        )
+        user_login_failed.connect(
+            user_login_failed_callback,
+            sender="django.contrib.auth",
+            dispatch_uid="user_login_failed",
+        )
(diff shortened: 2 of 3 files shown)


blueprintapi/settings.py
```diff
@@ -22,11 +23,21 @@
...
-        "handlers": ["console"],
-        "level": "WARNING",
+    "formatters": {
+        "simple": {
+            "format": "[{asctime}] [{levelname}] {message}",
+            "style": "{",
+        },
+    },
+    "loggers": {
+        "": {
```
jwhite3, line 37: Why not add a name for the logger here?

users/signals.py
```diff
@@ -1,25 +1,55 @@
...
+        component_perms = (
+            "add_component",
+            "change_component",
+            "view_component",
+        )
+        project_control_perms = (
+            "add_projectcontrol",
+            "change_projectcontrol",
+            "view_projectcontrol",
+        )
```
jwhite3, line 35: I'm guessing you ran `black`?

users/signals.py
```diff
@@ -1,25 +1,55 @@
...
             instance.user_permissions.add(permission)  # Model permissions
 
             if code in user_perms:
                 assign_perm(code, instance, instance)  # Object permissions for user
+
+
+@receiver(user_login_failed)
+def user_login_failed_callback(sender, credentials, **kwargs):
+    user = credentials.get("username")
+    logger.warning(f"Log in failed for: {user}")
```
jwhite3: Only define the receiver function here rather than using the decorator. Register and connect this signal over in `users.apps.py` (should be an example there already).

jwhite3 (changes requested):

blueprintapi/settings.py
TMCamp: If I give the formatter a name the format defined in `simple` doesn't apply to all of the loggers, but only to ones called by the name or that are using the `simple` formatter.

users/signals.py
TMCamp: 🥇 _sadly_

blueprintapi/settings.py
jwhite3: I see. In that case, why not apply it to `root`?

jauschalley (dismissed):
Looks great to me!

blueprintapi/settings.py
```diff
@@ -22,11 +23,21 @@
...
+        "simple": {
+            "format": "[{asctime}] [{levelname}] {message}",
+            "style": "{",
+        },
+    },
+    "loggers": {
+        "root": {
+            "level": "INFO",
+            "handlers": ["console"],
+        }
```
jwhite3: Sorry, this would apply to the logger called "root", but I'm not sure if that actually means that it will work for the root logger. Does it work as expected for all logs? If so, we can just clean this up later.

users/signals.py
```diff
@@ -1,25 +1,54 @@
...
 
             if code in user_perms:
                 assign_perm(code, instance, instance)  # Object permissions for user
+
+
+# noinspection PyUnusedLocal
+def user_login_failed_callback(sender, credentials, request, **kwargs):
+    user = credentials.get("username")
+    ip = request.META.get("REMOTE_ADDR")
+    logger.warning(f"Log in failed for {user} from IP {ip}")
```
jwhite3: Use lazy formatting here: 
```suggestion
    logger.warning("Log in failed for %s from IP %s", user, ip)
```

users/apps.py
```diff
@@ -1,11 +1,18 @@
...
-        from users.signals import add_default_user_perms
+        from users.signals import add_default_user_perms, user_login_failed_callback
 
-        post_save.connect(add_default_user_perms, sender="users.User", dispatch_uid="add_user_perms")
+        post_save.connect(
+            add_default_user_perms, sender="users.User", dispatch_uid="add_user_perms"
+        )
+        user_login_failed.connect(
+            user_login_failed_callback, dispatch_uid="failed_login"
+        )
```
jwhite3, line 18: Nice!

users/signals.py
```diff
@@ -1,25 +1,54 @@
...
+                break  # If the perms don't exist, likely a setup step, so don't try and add perms
 
             instance.user_permissions.add(permission)  # Model permissions
 
             if code in user_perms:
                 assign_perm(code, instance, instance)  # Object permissions for user
+
+
+# noinspection PyUnusedLocal
+def user_login_failed_callback(sender, credentials, request, **kwargs):
```
jwhite3: Not a blocker: might be nice to add some type hinting here so that we know what these args are supposed to be.

blueprintapi/settings.py
TMCamp: Ah, gotcha. Originally leaving the key in loggers blank came from here -> https://docs.djangoproject.com/en/4.1/howto/logging/#configure-a-logger-mapping

rapidato-github-svc (approved):

blueprintapi/settings.py
jwhite3: Oh, I see! I hadn't seen that before. Nice

jwhite3 (approved):

PR #96, opened 2022-08-30
Feature/catalog import am

jwhite3:
*What does this PR do?*
This PR adds 3.0 catalog JSON files to the repo and updates the `load_catalogs` command to include an option to "auto" load these catalogs.

Additionally, the `CSRF_TRUSTED_ORIGINS` was updated to allow access to the django admin page in deployed environments.

*Jira ticket number?*
[ISPGBSS-1235](https://jiraent.cms.gov/browse/ISPGBSS-1235)

*Changes*
- Add new option, `--load-standard-catalogs` that can be used to ingest catalogs included in the repo.
- Add new `Environment` attribute that corresponds to and is used by the `CSRF_TRUSTED_ORIGINS` setting


blueprintapi/Environment.py
```diff
@@ -13,7 +13,7 @@ def __init__(self):
             "django-insecure-_o$0y5g@1*uyrw0!3(0%wdv-ds5wp26yp*bko+q#y4b&y!50%6"
         )
         hosts = ["localhost", "127.0.0.1"]
-        env_hosts = os.environ.get("ALLOWED_HOSTS")
+        env_hosts = os.environ.get("ALLOWED_HOSTS", "localhost")
```
jwhite3, line 16: @azharem there were a few environment variables that I thought could use some reasonable defaults to make life easier when developing locally.

catalogs/data/README.md
```diff
@@ -0,0 +1,11 @@
...
+The following catalogs are sourced from https://github.com/CMSgov/ars-machine-readable .
+These catalogs are ingested into the blueprint database during deployment, and should not be modified. 
+Projects should be associated with a single catalog dataset given a catalog version and impact level 
+("high", "moderate", "low").
+
+## Version 3.1 (current)
+- CMS_ARS_3_1_[HIGH, MODERATE, LOW]_baseline_catalog.json
+
+## Version 5.0
+- cms_ars_50_[high, moderate, low].jon
```
jwhite3: Forgot to remove this. I'll drop this section. 

Let me know if more detail should be added here. @TMCamp might know what would be most useful here.

jwhite3:
@azharem the command for loading the catalogs will be:
```shell
python3 manage.py load_catalogs --load-standard-catalogs
```

blueprintapi/Environment.py
rapidato-github-svc: @jwhite3 The default "localhost" is already part of the Allowed_host. Look at line 15.

jwhite3: Yes, but before, `ALLOWED_HOSTS` was effectively not required, while now it is. So it has to be set to something since it's used in constructing `CSRF_TRUSTED_ORIGINS`, and I figured that "localhost" would be a reasonable default even though it meant that `ALLOWED_HOSTS` had a duplicate entry when the default is used.

catalogs/management/commands/load_catalog.py
```diff
@@ -16,25 +19,54 @@ def add_arguments(self, parser):
...
     def handle(self, *args, **options):
-        input_file = Path(options["catalog_file"])
+        if options["load_standard_catalogs"]:
+            self._load_standards()
+        else:
+            input_file = Path(options["catalog_file"])
+            name = input_name if (input_name := options["name"]) else input_file.name
+            create_kwargs = {}
+            for arg in ("source", "catalog_version", "impact_level"):
+                if value := options[arg]:
```
TMCamp, line 32: Isn't `:=` an assignment operator not a comparison operator? So you are assigning the value of `options[arg]` to value, but does this `if` fail if `arg` doesn't exist?

TMCamp (dismissed):

catalogs/management/commands/load_catalog.py
jwhite3: Order of operations: the assignment happens first, then the evaluation of the `if`, so effectively I'm checking whether or not `value` evaluates to `True` after it's been assigned. 

They're guaranteed to be in `options` since each has a default value.

blueprintapi/Environment.py
rapidato-github-svc: In that case I'd suggest to remove "localhost" from line 15 to make it like this: 

> hosts = ["127.0.0.1"]

jauschalley (dismissed):
This is exciting for so many reasons.  🚀

blueprintapi/Environment.py
jwhite3: I worried about the assumption that `ALLOWED_HOSTS` wouldn't be set to something else, which would mean that "localhost" would be missing from the allowed hosts django setting and could cause issues for local development. 

I compromised and switched to using a `set` to avoid duplicate entries in the allowed hosts list (which is probably a good idea anyways until we can switch to using the Django-environ plugin).

rapidato-github-svc (approved):

jauschalley (approved):

PR #99, opened 2022-09-02
Add test componenets

jwhite3:
*What does this PR do?*
This PR adds a management command that can be used to load a set of test components from the repository.

*Jira ticket number?*

*Changes*
- Added set of "test" components
- Added new management command, `load_components`

```shell
usage: manage.py load_components [-h] [--load-test-data LOAD_TEST_DATA]
                                 [--catalog CATALOG] [--version]
                                 [-v {0,1,2,3}] [--settings SETTINGS]
                                 [--pythonpath PYTHONPATH] [--traceback]
                                 [--no-color] [--force-color] [--skip-checks]

Ingest component data into the Component table. Currently assumes a single
component definition defined against a single catalog

optional arguments:
  -h, --help            show this help message and exit
  --load-test-data LOAD_TEST_DATA
  --catalog CATALOG
  --version             Show program's version number and exit.
  -v {0,1,2,3}, --verbosity {0,1,2,3}
                        Verbosity level; 0=minimal output, 1=normal output,
                        2=verbose output, 3=very verbose output
  --settings SETTINGS   The Python path to a settings module, e.g.
                        "myproject.settings.main". If this isn't provided, the
                        DJANGO_SETTINGS_MODULE environment variable will be
                        used.
  --pythonpath PYTHONPATH
                        A directory to add to the Python path, e.g.
                        "/home/djangoprojects/myproject".
  --traceback           Raise on CommandError exceptions.
  --no-color            Don't colorize the command output.
  --force-color         Force colorization of the command output.
  --skip-checks         Skip system checks.
```

### DIFF
--- a/components/management/commands/load_components.py
+++ b/components/management/commands/load_components.py
@@ -1,0 +1,64 @@
+import json
+
+from pathlib import Path
+
+from pydantic import ValidationError
+from django.core.management.base import BaseCommand, CommandError
+
+from blueprintapi.oscal.component import Model as ComponentModel
+from catalogs.models import Catalog
+from components.models import Component
+
+
+class Command(BaseCommand):
+    help = (
+        "Ingest component data into the Component table. "
+        "Currently assumes a single component definition defined against a single catalog"
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument("--load-test-data", type=bool, default=True)
+        parser.add_argument("--catalog", type=str, default="CMS_ARS_3_1_MODERATE")
+
+    def handle(self, *args, **options):
+        if options["load_test_data"]:
+            self._load_test_data(**options)
+
+    def _load_component(self, json_path: Path, catalog_name: str):
+        """Ingest a component from a valid OSCAL JSON file."""
+        try:
+            component = ComponentModel.from_json(json_path)
+        except ValidationError as exc:
+            raise CommandError(f"Could not load, {json_path.name} due to invalid data: {exc}") from exc
+
+        # Will need to update this to support loading multiple components from a single component file as a
+        # component may be defined relative to multiple catalogs in the future.
+        component_def = component.component_definition.components[0]
+
+        try:
+            catalog = Catalog.objects.get(name=catalog_name)
+        except Catalog.DoesNotExist as exc:
+            raise CommandError(f"Selected catalog {catalog_name} does not exist.") from exc
+
+        _, created = Component.objects.get_or_create(
+            title=component_def.title,
+            description=component_def.description,
+            type=component_def.type,
+            catalog=catalog,
+            controls=component_def.control_ids,
+            component_json=json.loads(component.json(by_alias=True, exclude_none=True)),
+            status=Component.Status.PUBLIC,
+        )
+
+        if created:
+            self.stdout.write(self.style.SUCCESS(f"Component, {component_def.title} was successfully created."))
+        else:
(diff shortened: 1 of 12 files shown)


components/management/commands/load_components.py
```diff
@@ -0,0 +1,64 @@
...
+        "Currently assumes a single component definition defined against a single catalog"
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument("--load-test-data", type=bool, default=True)
+        parser.add_argument("--catalog", type=str, default="CMS_ARS_3_1_MODERATE")
+
+    def handle(self, *args, **options):
+        if options["load_test_data"]:
+            self._load_test_data(**options)
```
jwhite3, line 25: I know this might look odd/thin, but the idea was that we'd might want to extend this to load components from another source or something. Even directly from the component repo.

components/management/commands/load_components.py
```diff
@@ -0,0 +1,64 @@
...
+
+class Command(BaseCommand):
+    help = (
+        "Ingest component data into the Component table. "
+        "Currently assumes a single component definition defined against a single catalog"
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument("--load-test-data", type=bool, default=True)
+        parser.add_argument("--catalog", type=str, default="CMS_ARS_3_1_MODERATE")
```
jwhite3, line 21: At some point I think this should be removed as an option and we should figure out what the catalog should be based on the component data.

components/tests.py
```diff
@@ -757,3 +758,29 @@ def test_add_control_description(self):
...
                 self.assertEqual(implemented.get("description"), test_description)
                 self.assertEqual(implemented.get("control-id"), test_control_id)
+
+
+class LoadComponentsTestCase(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        call_command("load_catalog", load_standard_catalogs=True)
+
+    def test_load_test_components(self):
```
jwhite3, line 768: Let me know if there are any other fields, etc that would be good to check.

components/tests.py
```diff
@@ -757,3 +758,29 @@ def test_add_control_description(self):
...
+            "Splunk",
+            "Tenable Nessus",
+            "Trend Micro Deep Security"
+        ]
+
+        queryset = Component.objects.order_by("title")
+
+        self.assertEqual(queryset.count(), 8)
+        self.assertEqual(expected_components, [item.title for item in queryset])
+        self.assertTrue(all(item.type == "software" for item in queryset))
```
TMCamp, line 785: OCISO type may change to `policy` going forward. Just a heads up.

TMCamp (approved):

components/tests.py
jwhite3: Only if we change the value in the file in this repo though ;). If we want to, we can of course.

components/management/commands/load_components.py
```diff
@@ -0,0 +1,64 @@
...
+        # component may be defined relative to multiple catalogs in the future.
+        component_def = component.component_definition.components[0]
+
+        try:
+            catalog = Catalog.objects.get(name=catalog_name)
+        except Catalog.DoesNotExist as exc:
+            raise CommandError(f"Selected catalog {catalog_name} does not exist.") from exc
+
+        _, created = Component.objects.get_or_create(
+            title=component_def.title,
```
jauschalley, line 44: I had to look up `_,` --- Python 🙉

jauschalley (approved):

PR #107, opened 2022-09-07
Adding SSP creator.

TMCamp:
*What does this PR do?*
Generates an OSCAL System Security Plan to be downloaded.

*Jira ticket number?*
[ISPGBSS-1099](https://jiraent.cms.gov/browse/ISPGBSS-1099)

*Changes*

- Added view and class for downloading an SSP

*Testing*

- [ ] If you haven't already, create a project
- [ ] In Postman, get an access token (localhost:8000/api-token-auth/)
- [ ] In Postman, submit a GET request to localhost:8000/api/projects/[PROJECT ID]/downloads/oscal
- [ ] You should get an OSCAL SSP JSON response.


blueprintapi/oscal/oscal.py
```diff
@@ -271,7 +271,7 @@ class RoleIDEnum(str, Enum):
 
 
 class Role(OSCALElement):
-    id: RoleIDEnum
+    id: str
```
jwhite3, line 274: So then is it no longer expected that the `id` will be one of `RoleIDEnum`?

projects/downloads.py
```diff
@@ -0,0 +1,158 @@
...
+    SystemImplementation,
+    SystemInformation,
+    SystemSecurityPlan,
+    SystemStatus,
+)
+from components.componentio import ComponentTools
+from projects.models import Project
+
+
+class OscalSSP: # pylint: disable=too-many-instance-attributes
```
jwhite3: Instead of disabling the warning, we should think about if there's a better approach. For example, there was a different class that was reduced to a single function in a previous PR: https://github.com/CMSgov/blueprint_api/pull/85/ (`EmptyComponent` -> `create_empty_component_json`). 

Not necessarily saying that it needs to be a function per se, but it does look like the main functionality/usage for this class will be `get_ssp`

projects/downloads.py
```diff
@@ -0,0 +1,158 @@
...
+    metadata = None
+    sec_impact_level = None
+    import_profile = None
+    information_type = None
+    system_information = None
+    system_characteristics = None
+    component_ref = {}
+    system_implementation = None
+    control_implementations = None
+    back_matter = None
```
jwhite3: Should all of these be class attributes? Instances of this class will share these values, so was that intentional?

projects/downloads.py
jwhite3: In fact, I wonder if this would be better as a method on the `Project` model since it's effectively a transformation of a project instance.

blueprintapi/oscal/oscal.py
TMCamp: It turns out that Roles defined in that enum are FedRAMP roles, but not necessarily universally specific.

From the OSCAL Gitter channel:
`Since <role> lacks any means of association (has no suitable attribute) with a well-known framework (such as NIST RMF roles defined in NIST SP 800-37 Appendix D) the role identifier is essentially unconstrained.`

projects/downloads.py
TMCamp: No. That was a leftover from some tinkering that I forgot to undo.

blueprintapi/oscal/ssp.py
```diff
@@ -516,85 +516,3 @@ def json(self, **kwargs):
             kwargs.pop("exclude_none")
 
         return super().json(by_alias=True, exclude_none=True, **kwargs)
-
-
-def main():
```
openprivacy, line 521: Are there any tests for this class? It would be useful to have this turned into a test...

jwhite3:
Need to hold off on this PR until #113 is done (it touches quite a lot, and is a critical fix) I'm going to set this as a draft until that one is done. 

@TMCamp

projects/project_extra.json
```diff
@@ -0,0 +1,60 @@
+{
```
jwhite3, line 1: @TMCamp Is this what you were talking about needing additional input from the user?

TMCamp: That is one of the things. We need Roles, and we were able to pull those roles form CFACTS, but I'd rather not hardcode them into the download.

jauschalley (dismissed):
LGTM

projects/project_extra.json
```diff
@@ -0,0 +1,60 @@
...
+      }
+    },
+    {"System Read Only (SRO)":  {
+        "props": {
+          "name": "type",
+          "value": "internal"
+        }
+      }
+    },
+    {"System Read Only (SRO)":  {
```
openprivacy, line 24: Is SRO intended to be repeated?

openprivacy (dismissed):
LGTM2

projects/project_extra.json
jwhite3: But isn't that effectively what this is doing? Was there an issue in allowing for user input, or did you want to save that for a future iteration or something?

jwhite3:
@TMCamp would you mind updating this branch with dev? There are a few changes here that I think will need to be addressed once it's updated with dev since there were changes to the component and catalog tables.

components/serializers.py
```diff
@@ -27,6 +26,8 @@ class Meta:
             "description",
             "type",
             "supported_catalog_versions",
+            "catalog",
```
jwhite3: Need to update with dev. This will break since the component model in dev no longer has a `catalog` field

jwhite3 (changes requested):

jwhite3:
Also, I'm not seeing any tests for this new feature anywhere, was this actually still in progress?

projects/project_extra.json
jwhite3: @TMCamp Why mark this as resolved? The question wasn't answered.

jwhite3 (commented):
Conversation still isn't resolved, and we still aren't testing the new view/endpoint anywhere.

projects/project_extra.json
TMCamp: Asking the user to input the roles that they have listed in CFACTS is not really acceptable because we are asking the user to maintain data in separate systems and increasing the risk for errors. Until we can import the Project data in toto this should work because it adds roles that are common across most CMS projects.

jwhite3: Ok, I think I understand that, but why the JSON file? It's so small that it seems like it would be far more useful to define as a dictionary in python rather than reading it in on each of these requests. 

Even further, why parse this structure at all if they're meant as defaults (effectively)? Is the parsing meant to be something that would also parse future user-supplied data?

TMCamp: Yes, I thought that it would be easier to parse future user-supplied data and/or one could just replace the file to add/change defaults rather than updating hardcoded values.

jwhite3 (dismissed):

jauschalley (dismissed):
🚢 it

jwhite3 (dismissed):

jauschalley (dismissed):

jwhite3 (dismissed):

jauschalley (dismissed):

projects/serializers.py
```diff
@@ -92,12 +91,7 @@ class ProjectControlSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = ProjectControl
-        fields = ("status", "project", "control", "remarks", "catalog_data", "component_data", )
-
-    def validate(self, attrs):
-        if attrs.get("status") == ProjectControl.Status.NA and not attrs.get("remarks"):
-            raise serializers.ValidationError("A justification is required for non-applicable controls.")
-        return attrs
```
jwhite3: Was this intentional?

TMCamp: No!

projects/serializers.py
```diff
@@ -179,6 +178,5 @@ class Meta:
             "status",
             "control",
             "project",
-            "remarks",
```
jwhite3, line 224: This as well; was it intentional to remove this?

TMCamp: Got it

jwhite3 (dismissed):

PR #110, opened 2022-09-08
Update and sync python version to 3.10

jwhite3:
*What does this PR do?*
This PR updates the python version to 3.10 and syncs all Actions workflows and the Dockerfile to use the same version. Additionally, we define a more strict python version requirement in `pyproject.toml`

*Jira ticket number?*
[ISPGBSS-1243](https://jiraent.cms.gov/browse/ISPGBSS-1243)

*Changes*
- Update python version to "3.10" in Actions workflows
- Update python version requirement in `pyproject.toml`
- Update python docker image

### DIFF
--- a/blueprintapi/environment.py
+++ b/blueprintapi/environment.py
@@ -21,7 +21,7 @@ class Environment:
             # allowed hosts.
             metadata_uri = os.environ.get("ECS_CONTAINER_METADATA_URI")
             if metadata_uri:
-                container_metadata = requests.get(metadata_uri).json()
+                container_metadata = requests.get(metadata_uri, timeout=30).json()
                 hosts.add(container_metadata["Networks"][0]["IPv4Addresses"][0])
         cors_origin = (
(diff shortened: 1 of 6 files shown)


blueprintapi/environment.py
```diff
@@ -21,7 +21,7 @@ def __init__(self):
             # allowed hosts.
             metadata_uri = os.environ.get("ECS_CONTAINER_METADATA_URI")
             if metadata_uri:
-                container_metadata = requests.get(metadata_uri).json()
+                container_metadata = requests.get(metadata_uri, timeout=30).json()
```
amyd11, line 24: Just curious, what's the timeout for here?

amyd11 (approved):

blueprintapi/environment.py
jwhite3: `pylint` suggestion, but it's good practice regardless in case something hangs. We don't want an endless loop!

jwhite3: 30 -> 30s btw

TMCamp (approved):

PR #111, opened 2022-09-08
Add session auth

jwhite3:
*What does this PR do?*
This PR adds session authentication to the api. This allows for users to login with the rest framework UI and swagger UI and make requests.

*Jira ticket number?*

*Changes*
- Add SessionAuthentication class to default classes
- Add swagger settings for login/logout pages

### DIFF
--- a/blueprintapi/settings.py
+++ b/blueprintapi/settings.py
@@ -117,7 +117,6 @@ AUTHENTICATION_BACKENDS = (
     "guardian.backends.ObjectPermissionBackend",
 )
 
-
 # Database
 # https://docs.djangoproject.com/en/3.2/ref/settings/#databases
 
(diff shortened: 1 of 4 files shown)


blueprintapi/settings.py
```diff
@@ -201,10 +197,16 @@
...
+        "rest_framework.authentication.SessionAuthentication",
     ],
     "DEFAULT_PERMISSION_CLASSES": [
         "blueprintapi.permissions.StrictDjangoObjectPermissions",
     ],
     "DEFAULT_SCHEMA_CLASS": "rest_framework.schemas.coreapi.AutoSchema",
     "PAGE_SIZE": 20,
 }
+
+SWAGGER_SETTINGS = {
```
amyd11, line 209: Is this being referenced somewhere or is there some magic that happens behind the scenes?

jwhite3: More or less magic. Similar to how you can set global settings with rest framework, `drf-yasg` also has some global settings that you can configure.

amyd11 (approved):

jauschalley (approved):
Oh yeah! This is awesome.  Thanks!

PR #113, opened 2022-09-12
1257 decouple components catalogs

jwhite3:
*What does this PR do?*
This PR removes the foreign key relationship between a component instance and catalog instance since a component can be associated with more than one catalog (of a supported version).

*Jira ticket number?*
[ISPGBSS-1257](https://jiraent.cms.gov/browse/ISPGBSS-1257)

*Changes*
- Remove foreign key relationship between Component and Catalog
- Update serializers, views, and tests
- Add new pre-save signal to pull out supported versions
- Update load commands


jauschalley (commented):
Looking good.  Let me know if you want me to pull it down and poke at it.

jwhite3:
> Looking good. Let me know if you want me to pull it down and poke at it.

Yeah definitely. I'm going to do the same, but It's always nice to have a second pair of eyes/hands

catalogs/models.py
```diff
@@ -55,6 +60,9 @@ class ImpactLevel(models.TextChoices):
     def __str__(self):
         return self.name
 
+    class Meta:
+        unique_together = ("version", "impact_level")
+
```
TMCamp, line 65: Nice!

catalogs/catalogio.py
```diff
@@ -126,11 +126,13 @@ def get_next_control_by_id(self, control_id: str):
...
-            raise MissingControlError(
-                "Cannot determine next control. Provided id does not exist"
-            ) from exc
+        except ValueError:
+            logger.warning(
+                "Cannot determine next control. Provided control does not exist in catalog(s): %s", control_id
+            )
+            return ""
         except IndexError:  # control_id was at the end of the list
+            logger.info("No new controls.")
```
jwhite3, line 135: Apparently it's allowed for components to be defined with controls that might not be in the catalogs..

blueprintapi/oscal/component.py
```diff
@@ -161,6 +162,17 @@ def control_ids(self) -> List[str]:
             }
         )
 
+    @property
+    def catalog_versions(self) -> List[Catalog.Version]:
+        versions = set()
+        for item in self.control_implementations:
+            try:
+                versions.add(Catalog.Version(item.description))
```
amyd11, line 189: Yay, ok cool- this is where we're creating the "real" list of catalogs that a component belongs to

catalogs/management/commands/load_catalog.py
```diff
@@ -41,7 +41,7 @@ def _load_catalog(self, input_file: Path, name: str, **catalog_args):
 
         try:
             with open(input_file, mode="rb") as catalog:
-                file = File(catalog)
+                file = File(catalog, name=input_file)
```
amyd11, line 44: This looks like you fixed a bug!

components/filters.py
```diff
@@ -10,11 +10,11 @@ class ComponentFilter(filters.FilterSet):
 
     search = filters.CharFilter(method="keyword_search", label="Search")
     type = filters.CharFilter(lookup_expr="iexact")
-    catalog = filters.NumberFilter()
+    catalog_version = filters.CharFilter(field_name="supported_catalog_versions", lookup_expr="contains")
 
     class Meta:
         model = Component
-        fields = ["search", "type", "catalog"]
+        fields = ["search", "type"]
```
amyd11: This feels better.

catalogs/management/commands/load_catalog.py
jwhite3: It's... something. Technically the thing I was trying to fix is still kind of there, but just got booted. Django does some stuff to "media" fields that results in weird behavior that I don't think that we really need in our case since we're effectively loading a "static" dataset from the repo. I was torn between trying to make it work vs updating that field to _not_ be a "media" field. Wasn't sure I could justify this one as "in-scope". 

Effectively though, I don't think it'll make too much of a difference in the deployed environments, but for local development, if you delete your local "media" folder every now and again you may get a `FileNotFound` error because of the magic django tries to do in the background to protect the "original" files (effectively makes copies and references the copies in certain circumstances).

components/filters.py
jwhite3: Note that the query key is now `catalog_version` instead of `catalog` too.

catalogs/catalogio.py
TMCamp: When we generate the SSP or show the Controls on a Project page, the Controls come from the associated Catalog, so if a Component has Controls that aren't pertinent, they will not be applied. In a similar vein a Component can reference Controls that are referenced for Low, Moderate _and_ High impact level. If the Component is being applied to a Low system, the Controls that only pertain to Moderate and High will be ignored.

TMCamp (dismissed):

jauschalley (dismissed):

components/serializers.py
```diff
@@ -110,25 +109,26 @@ def collect_catalog_data(controls: list, catalog_versions: List[Catalog.Version]
...
+                "source": item.source,
+                "controls": {
+                    control.control_id: {
+                        "narrative": control.description,
+                        "responsibility": control.responsibility,
+                        "provider": control.provider
+                    }
+                    for control in item.implemented_requirements
+                }
+            }
```
amyd11, line 129: Thank you for this- the similar structure to the catalog data that comes back makes it much easier to understand when working with the data

amyd11 (dismissed):

amyd11:
I should have officially noted: I've been testing this as I've been building out the corresponding front end changes ([UI PR](https://github.com/CMSgov/blueprint_ui/pull/95)) and it's been working great!

amyd11 (approved):

jauschalley (approved):

PR #115, opened 2022-09-14
Fix controls ordering

jwhite3:
*What does this PR do?*
Fix sorting in identifying "next" control, and fix controls view queryset to include an order by.

*Jira ticket number?*
[ISPGBSS-1261](https://jiraent.cms.gov/browse/ISPGBSS-1261)

*Changes*
- Order controls in `catalogio` by compound key to correctly order by id
- Order queryset for `ProjectGetControlList`

### DIFF
--- a/catalogs/catalogio.py
+++ b/catalogs/catalogio.py
@@ -1,6 +1,6 @@
 import json
 import logging
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Tuple
 
 logger = logging.getLogger("catalogs.catalogio")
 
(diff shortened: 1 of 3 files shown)


TMCamp (dismissed):

jauschalley:
This makes sense to me though I may be missing some depth from the issue that is being experienced.

jwhite3:
> This makes sense to me though I may be missing some depth from the issue that is being experienced.

I think there were two bits: 
1. The response from `/api/projects/.../controls/` was unordered, which isn't great for a paginated dataset regardless and may have been introducing unexpected behavior
2. The "next_id" field wasn't providing what you'd expect to be "next". For example, for "ac-1" you'd expect "ac-2", but this wasn't necessarily the case for all ids since the sorting was a simple string sort. You'd get "ac-10" or "ac-14" before "ac-2"

jwhite3:
Setting to draft until #113 is done.

jauschalley (approved):
Seems straight forward to me.

amyd11 (approved):
Just tested it out & it works great. I'm excited about this fix!

PR #116, opened 2022-09-14
Feature/1236 non applicable

TMCamp:
*What does this PR do?*

*Jira ticket number?*
[ISPGBSS-1236](https://jiraent.cms.gov/browse/ISPGBSS-1236)

*Changes*

- Add `not-applicable` to the _Status_ choices
- Add `remarks` column to the _ProjectControl_ table

*Testing*

- Assuming you have created a Project:
- [ ] in Postman create a Patch request to localhost:8000/api/projects/[PROJECT ID]/controls/ac-1/
- [ ] In the Body, add Form-fields `status`: `not_applicable`, and `remarks`: `This is the remarks`
- [ ] Send the Request
- [ ] The response should show the updated status and the remark.


projects/serializers.py
```diff
@@ -86,7 +86,7 @@ class ProjectControlSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = ProjectControl
-        fields = ("status", "project", "control", "catalog_data", "component_data", )
+        fields = ("status", "project", "control", "remarks", "catalog_data", "component_data", )
```
jwhite3, line 95: Should probably add validation somewhere so that if someone tries to set "status" to NA, but doesn't provide "remarks", that they'll get a 400. I think that was a requirement right (that you _had_ to provide a reason if NA)?

jwhite3:
Setting this to a draft until #113 is done.

projects/serializers.py
TMCamp: That is a requirement, but I hadn't thought of enforcing it on the back end. Good call!

jwhite3:
> Setting this to a draft until #113 is done.

^^^

projects/models.py
```diff
@@ -118,6 +119,7 @@ class Status(models.TextChoices):
         null=False,
         help_text="The Project Control status; completed, incomplete, or not started",
     )
+    remarks = models.TextField(blank=True)
```
jwhite3, line 122: Just curious: why not "comment"? (not a required change or anything)

projects/serializers.py
```diff
@@ -86,7 +86,12 @@ class ProjectControlSerializer(serializers.ModelSerializer):
...
+
+    def validate(self, attrs):
+        if attrs.get("status") == ProjectControl.Status.NA and attrs.get("remarks") is None:
+            raise serializers.ValidationError("A justification is required for non-applicable controls.")
```
jwhite3: Need to also check that "remarks" isn't empty:
```suggestion
        if attrs.get("status") == ProjectControl.Status.NA and not attrs.get("remarks"):
            raise serializers.ValidationError("A justification is required for non-applicable controls.")
```

jwhite3 (changes requested):
Just the one small change to the validation.

projects/models.py
TMCamp: In OSCAL they are Remarks so I was trying to keep it consistent.

projects/serializers.py
jwhite3: @TMCamp Why not use this? Using `not attrs.get("status")` more concisely covers both the case where the value is an empty string or `None`.

projects/serializers.py
```diff
@@ -86,7 +86,13 @@ class ProjectControlSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = ProjectControl
-        fields = ("status", "project", "control", "catalog_data", "component_data", )
+        fields = ("status", "project", "control", "remarks", "catalog_data", "component_data", )
+
+    def validate(self, attrs):
+        remarks = attrs.get("remarks")
+        if attrs.get("status") == ProjectControl.Status.NA and (remarks is None or remarks == ""):
+            raise serializers.ValidationError("A justification is required for non-applicable controls.")
```
jwhite3: Not sure where my original comment went, but I still don't understand why this was implemented in this way. I had suggested a more concise check:

```suggestion
        if attrs.get("status") == ProjectControl.Status.NA and not attrs.get("remarks"):
            raise serializers.ValidationError("A justification is required for non-applicable controls.")
```

Can you explain your reasoning?

projects/serializers.py
jwhite3: Found it. This is now a duplicate of a new conversation.

projects/serializers.py
TMCamp: not sure, but I change it.

jwhite3 (dismissed):

jwhite3 (dismissed):

jauschalley (dismissed):

jwhite3 (approved):

jauschalley (approved):

PR #117, opened 2022-09-19
Fix swagger request schemes

jwhite3:
*What does this PR do?*
This PR fixes allowed schemes for the swagger docs ui.

*Changes*
- Add manual schemes list

### DIFF
--- a/blueprintapi/urls.py
+++ b/blueprintapi/urls.py
@@ -16,10 +16,19 @@ Including another URLconf
 from django.contrib import admin
 from django.urls import include, path, re_path
 from drf_yasg import openapi
+from drf_yasg.generators import OpenAPISchemaGenerator
 from drf_yasg.views import get_schema_view
 from rest_framework import permissions
 
 from . import views
+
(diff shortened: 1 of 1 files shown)


amyd11 (approved):
Code changes make sense. This seems like something that's hard to test locally- I think it's safe to merge then test (and we can always continue to fix if necessary!)

jwhite3:
> Code changes make sense. This seems like something that's hard to test locally- I think it's safe to merge then test (and we can always continue to fix if necessary!)

Yeah, I had no idea how to actually do that in a test :/

jauschalley:
> > Code changes make sense. This seems like something that's hard to test locally- I think it's safe to merge then test (and we can always continue to fix if necessary!)
> 
> Yeah, I had no idea how to actually do that in a test :/

I think there is a way to lock down docker to HTTPS between containers but I'm really not sure how to do this within the Django app itself.

jauschalley (approved):
Thanks!

PR #120, opened 2022-09-22
Feature/add 5.0 catalogs

jwhite3:
*What does this PR do?*
Add support to ingest 5.0 catalogs

*Jira ticket number?*
[ISPGBSS-1268](https://jiraent.cms.gov/browse/ISPGBSS-1268)

*Changes*
- Add version 5.0 pydantic models
- Update control signal
- Add 5.0 standards
- Update catalogs tests


jwhite3:
To do: 
- [x] Add additional testing
- [x] Add 5.0 catalogs testing coverage

catalogs/io/v5_0.py
```diff
@@ -0,0 +1,193 @@
...
+class BaseControl(BaseModel):
+    class Meta:
+        allow_population_by_field_name = True
+
+    id: str
+    item_class: str = Field(..., alias="class")
+    title: str
+
+
+class Prop(BaseModel):
```
TMCamp, line 33: This is in `blueprintapi.oscal.oscal.Property`

catalogs/io/v5_0.py
```diff
@@ -0,0 +1,193 @@
...
+    name: str
+    value: str
+
+
+class Link(BaseModel):
```
TMCamp, line 38: This is in `blueprintapi.oscal.oscal.Link`

catalogs/io/v5_0.py
```diff
@@ -0,0 +1,193 @@
...
+    def set_family_id(cls, value: list[Control], values) -> list[Control]:  # pylint: disable=no-self-argument
+        for item in value:
+            if not item.family_id:
+                item.family_id = values.get("id")
+
+        return value
+
+
+class CatalogModel(BaseModel):
+    uuid: UUID4
```
TMCamp, line 151: You could use `uid: UUID = Field(default_factory=uuid4)`

catalogs/io/v5_0.py
```diff
@@ -0,0 +1,193 @@
...
+
+    @classmethod
+    def from_json(cls, json_file: Union[str, Path]):
+        with open(json_file, "rb") as file:
+            data = json.load(file)
+
+        try:
+            return cls(**data)
+        except ValidationError:  # Try nested "catalog" field
+            return cls(**data["catalog"])
```
TMCamp, line 193: We might want to consider changing `catalogs/io/v5_0.py` to `blueprintapi/oscal/catalog.py` since the OSCAL Catalog model is not ARS version specific and should work with 3.1 Catalogs as well as with 800-53r5 catalog. We may also consider pulling the `oscal` directory out of the `blueprintapi` directory and making it a standalone lib.

catalogs/signals.py
```diff
@@ -3,28 +3,35 @@
...
-            for control_id in control_ids:
-                control_data = catalog.get_control_data_simplified(control_id)
-                control_objects.append(
-                    Controls(
-                        catalog=instance,
-                        control_id=control_id,
-                        control_label=control_data.get("label"),
-                        sort_id=control_data.get("sort_id"),
-                        title=control_data.get("title"),
+        if instance.version == Catalog.Version.CMS_ARS_3_1:
```
TMCamp, line 13: This conditional check _may_ not be necessary.

TMCamp (changes requested):

catalogs/io/v5_0.py
jwhite3: Why? Pydantic defines a type that does this for you and has less boilerplate

catalogs/signals.py
jwhite3: I think it is. I tried using the model I defined for 5.0 specifically on the 3.1 catalogs and there were enough differences that the models failed to load.

catalogs/io/v5_0.py
jwhite3: The only difference that I can think of is that it'll generate a uuid if the filed isn't present in the input data, but I'm not sure why we'd want to generate a new uuid if the field was missing?

jwhite3:
@TMCamp what do you think about consolidating the models between this implementation for the 5.0 catalogs and the existing models in `blueprintapi.oscal.oscal` in a followup PR?

catalogs/io/v5_0.py
jwhite3: In terms of separating out the catalog stuff into it's own lib, I suppose that's an option, but if we're going to keep catalog parsing code in this repo, I think it makes the most since to actually consolidate all things catalog related in the `catalogs` app within this project.

catalogs/io/v5_0.py
jwhite3: Going to tackle this in a separate PR

catalogs/io/v5_0.py
jwhite3: Going to tackle this in a separate PR

catalogs/io/v5_0.py
jwhite3: Hashed this out in a conversation. Going to leave as is

catalogs/io/v5_0.py
jwhite3: Talked about this a bit more "in person". The idea of consolidating "io" code with the catalog, component, etc repos is a great idea, but outside the scope of blueprint. In the future, blueprint could use that rather than implementing its own parsing code and schemas. 

Consolidating the models will happen in a separate, followup PR.

catalogs/signals.py
jwhite3: This check is necessary for now, but may become unnecessary after consolidating the models (which will happen later in a followup PR)

jauschalley (dismissed):

TMCamp:
> @TMCamp what do you think about consolidating the models between this implementation for the 5.0 catalogs and the existing models in `blueprintapi.oscal.oscal` in a followup PR?

Yep, sounds good to me.

TMCamp (dismissed):

amyd11 (approved):

TMCamp (approved):

jauschalley (approved):

PR #124, opened 2022-09-26
1018 disable narratives

jwhite3:
*What does this PR do?*
This PR implements the ability to enable/disable narratives on a specific control for a project.

*Jira ticket number?*
[ISPGBSS-1018](https://jiraent.cms.gov/browse/ISPGBSS-1018)

*Changes*
- Add new field, `disabled_narratives` to `ProjectControl` model
- Update serializer with new actions `disable_narratives` and `enable_narratives`


projects/serializers.py
```diff
@@ -133,13 +163,19 @@ def _get_control_data(components: QuerySet, control_id: str, catalog_version: st
...
+                    }
                 else:
                     # noinspection PyTypeChecker
                     result["components"][status][component.title] = {
+                        "id": component.id,
                         "description": control_data.description,
                         "responsibility": responsibility,
-                        "provider": control_data.provider
+                        "provider": control_data.provider,
+                        "enabled": enabled
```
jwhite3, line 203: New field in the component data will be "enabled" and it'll be a Boolean. Should be True by default since the `disabled_narratives` field will be an empty list by default.

projects/serializers.py
```diff
@@ -85,10 +84,39 @@ class ProjectControlSerializer(serializers.ModelSerializer):
...
+        fields = (
+            "status", "project", "control", "catalog_data", "component_data", "disable_narratives", "enable_narratives"
+        )
+
+    def update(self, instance, validated_data):
+        if disable := validated_data.get("disable_narratives"):
+            instance.disabled_narratives = list(set(instance.disabled_narratives).union(disable))
+
+        if enable := validated_data.get("enable_narratives"):
+            instance.disabled_narratives = list(set(instance.disabled_narratives).difference(enable))
```
jwhite3, line 127: Set union to avoid adding duplicate values. Need to validate the input data so that only allowed values can be added (and to avoid this list growing arbitrarily large since it's unconstrained against the list of project.components.

Set difference removes only those items actually contained in the original set and disregards extras.

jwhite3:
To do: 
- [x] Validate `disable_narratives` list against available project component ids
- [x] Add tests

projects/serializers.py
```diff
@@ -85,10 +84,39 @@ class ProjectControlSerializer(serializers.ModelSerializer):
     project = BasicViewProjectSerializer(read_only=True)
     catalog_data = serializers.SerializerMethodField(read_only=True)
     component_data = serializers.SerializerMethodField(read_only=True)
+    disable_narratives = serializers.ListSerializer(child=serializers.IntegerField(), write_only=True, required=False)
+    enable_narratives = serializers.ListSerializer(child=serializers.IntegerField(), write_only=True, required=False)
```
jwhite3, line 91: @jauschalley New fields `disable_narratives` and `enable_narratives`. Should be a list, but if only updating one thing at a time it can just be a list of one.

jwhite3: patch example:
```json
{
  "disable_narratives": [1, 2, 3],
  "enable_narratives": [4]
}
```

TMCamp (dismissed):

amyd11 (approved):

jauschalley (approved):

jauschalley:
Merging before we lose access....

